Issue #3, opened 2017-06-26
Delete / rename branches


shana:
Add support for deleting and renaming branches.

Basic:
- [x] Delete local branches
- [ ] Rename local branches

Advanced:
- [x] Delete remote branches. This is fairly simple to implement but needs heavy warnings and we need to make sure people don't shoot themselves in the foot with it.

PR #4, opened 2017-06-26
Add a UI for deleting a branch

shana:
As requested in #3
This pull request adds a UI for deleting a branch:

delete

> @donokuda wrote:
> ❓ I'm having issues having the button start off as disabled. Right now I'm comparing if `activeBranchNode` and `selectedNode` are the same, which according to `Debug.Log` seems to return the correct value. However, it's doesn't seem to actually work as intended.
> 
> Any insights would be nice!


StanleyGoldman:
Is this ready? i'm not sure...

StanleyGoldman:
@shana Is this good to go?

shana:
Yup, good to go!

PR #6, opened 2017-06-26
Add a UI for publishing repos

### LINKED ISSUE #5: Publish project to GitHub
When a project is on a local git repository but has no remote set up, we need to have a `Publish` UI to allow publishing repositories to a GitHub server without having to leave Unity.

- [ ] Detect if user is authenticated to control presenting "Publish" button
- [ ] Detect whether repository has been published (does it have a remote?)
- [ ] Hide Push/Pull buttons and show a Publish button (history view? branches view? everywhere?)
- [ ] Have a publish window that gathers the basic information for publishing a repo
  - Owner (either org or user), Name, Description (optional)
  - Make the repo public or private?

Validations:
- Is the user logged in? If not, we'll need to fire up the authentication flow before being able to publish
- Repository name
  - [ ] Name duplication - we can check whether the wanted name already exists with octokit
  - [ ] Spaces should be turned into dashes
- Private repository
  - [ ] If the user can't create private repositories, we need to show a useful error message

Advanced:
- [ ] Add support for selecting a license

UI Questions:
- The header should reflect whether the repo is published or not. Should we show the url there if it is?

shana:
Fixes #5 

Depends on:
- [x] #87 

> @donokuda wrote:

When a local repo doesn't exist on GitHub, we should give the user an opportunity to publish it from Unity so that they can start sharing the project with other teammates.

This pull request implements the UI for publishing a repo (but still needs hooking up):

**FUN GIF TIME!** 🎥 ✨ 

❓ When the user isn't logged in, should we pop open the Authentication window instead so that they can publish to GitHub? If we do, is there a way to "remember" that they were in the middle of publishing and replace the window with the Publish window?


StanleyGoldman:
It was suggested at one point, that we try to always use the [Repos - Create](https://developer.github.com/v3/repos/#create) method that takes an organization and supply the login username instead of an organization. The hope was that would help secure us from accidental publishing to a new user when we were expecting the previously authenticated user.

I did some tests and I see that is not possible.

Issue #64, opened 2017-06-29
ReadOnlyCollection type conflicts with VRTK


shana:
Looks like our inclusion of the ReadOnlyCollectionInterfaces nuget (to get `System.Collections.Generic.ReadOnlyCollection<T>` with the correct inheritance tree that satisfies Octokit's needs) is clashing with other plugins' use of the old `System.Collections.ObjectModel.ReadOnlyCollection<T>` type.

The easy fix is for VRTK to disambiguate the type by specifying the namespace to it. However, we are adding new types to the .net namespace that scripts aren't expecting, so this breakage is likely to happen to other scripts as well. We should probably be nice citizens and not add types to the System namespaces to avoid random clashes like this.

We might have to remove or change `ReadOnlyCollection<T>` usage from Octokit so we don't have to depend on the new type, or we can rename the namespace so it's not sitting in the System.* namespace.

![image](https://user-images.githubusercontent.com/310137/27701126-3f980410-5d00-11e7-8cbf-23702ebc9896.png)

shana:
I just published a new release with a fix for this https://github.com/github-for-unity/Unity/releases/tag/v0.13-alpha

PR #65, opened 2017-06-29
Moving ReadOnlyCollection to a different namespace

StanleyGoldman:
Fixes: #64 
Fixes: #69

- For ease of development we have removed submodules and included them as compiled DLLs.
- Added a solution `GitHub.Unity.Internal` that expects to find dependent repos as siblings to this repo, their names should be the same as their repositories
- Added functionality for the projects check the solution name, if it is equal to `GitHub.Unity.Internal`  they set a flag
- If the flag is set dependent projects are included with project references instead of libraries


src/GitHub.Api/GitHub.Api.csproj
```diff
@@ -11,6 +11,7 @@
     <AssemblyName>GitHub.Api</AssemblyName>
     <TargetFrameworkVersion>v3.5</TargetFrameworkVersion>
     <FileAlignment>512</FileAlignment>
+    <SolutionType Condition="$(SolutionName)=='GitHub.Unity.Internal'">Internal</SolutionType>
```
StanleyGoldman: Part one of the trick

src/GitHub.Api/GitHub.Api.csproj
```diff
@@ -243,19 +244,38 @@
       <Link>Properties\ApplicationInfo_Local.cs</Link>
     </Compile>
   </ItemGroup>
+  <Choose>
+    <When Condition="$(SolutionType) == 'Internal'">
+      <ItemGroup>
```
StanleyGoldman: Part two of the trick

shana:
Renamed the solution to `GitHub.Unity.OctokitDebugging.sln` to be more clear as to what it is.

Replaced the octokit and threading dlls with mono-built ones to get the symbol files when running with Unity.

Moved the custom targets out to separate files so VS doesn't keep screwing up the formatting.

Issue #73, opened 2017-06-30
Implement Git Fetch


StanleyGoldman:

thorade:
Should there be a notification upon success?
Something like: Already up to date? Analogous to the Pull button?

StanleyGoldman:
We talked about it [here](https://github.com/github-for-unity/Unity/pull/74#pullrequestreview-48048740). We came to the decision that we dont need the confirmation since `git fetch` does not change anything in your workspace directly.

thorade:
Yes, I saw that. I think a confirmation ("Do you REALLY want to fetch?") is not necessary, but some feedback that fetching finished successfully might be helpful!?
Can you use the status bar at the very bottom of the window maybe?

StanleyGoldman:
Using the status bar may not be a bad idea. Do me a favor and throw a feature into the backlog and we can talk about it.

thorade:
Sorry, what do you mean by backlog? Open a new issue?

StanleyGoldman:
Yea, sorry, that's what i meant.

PR #74, opened 2017-06-30
Adding git fetch button

StanleyGoldman:
Fixes: #73


src/UnityExtension/Assets/Editor/GitHub.Unity/UI/HistoryView.cs
```diff
@@ -321,8 +328,19 @@ public void OnEmbeddedGUI()
 
                 GUILayout.FlexibleSpace();
 
+                GUI.enabled = currentRemote != null;
+                var fetchClicked = GUILayout.Button(FetchButtonText, Styles.HistoryToolbarButtonStyle);
+                GUI.enabled = true;
+                if (fetchClicked &&
+                    EditorUtility.DisplayDialog(FetchConfirmTitle,
```
shana: We probably don't need to ask for confirmation, fetch is a non-destructive operation, we can just do it.

PR #91, opened 2017-07-03
Updating display correctly when the remote is removed

StanleyGoldman:
Fixes #89
Fixes #94 
Fixes #106 

Depends on:
- [x] #152
- [x] #153
- [x] #154
- [x] #155
- [x] #156
- [x] #157
- [x] #158
- [x] #159
- [x] #160
- [x] #161
- [x] #162
- [x] #163
- [x] #164
- [x] #165
- [x] #166
- [x] #167

PR #95, opened 2017-07-05
Fix to update header when remote is removed

StanleyGoldman:
Depends on:
- [ ] #91

### DIFF
--- a/src/UnityExtension/Assets/Editor/GitHub.Unity/UI/HistoryView.cs
+++ b/src/UnityExtension/Assets/Editor/GitHub.Unity/UI/HistoryView.cs
@@ -30,6 +30,13 @@ namespace GitHub.Unity
         private const string ClearSelectionButton = "×";
         private const string NoRepoTitle = "No Git repository found for this project";
         private const string NoRepoDescription = "Initialize a Git repository to track changes and collaborate with others.";
+        private const string FetchActionTitle = "Fetch Changes";
+        private const string FetchButtonText = "Fetch";
+        private const string FetchFailureDescription = "Could not fetch changes";
+        private const string FetchConfirmTitle = "Fetch Changes?";
+        private const string FetchConfirmDescription = "Would you like to fetch changes from remote '{0}'?";
+        private const string FetchConfirmYes = "Fetch";
+        private const string FetchConfirmCancel = "Cancel";
         private const int HistoryExtraItemCount = 10;
         private const float MaxChangelistHeightRatio = .2f;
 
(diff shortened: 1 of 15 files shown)


src/tests/IntegrationTests/Events/RepositoryManagerTests.cs
```diff
@@ -294,6 +294,9 @@ await RepositoryManager
             RepositoryManager.ActiveRemote.Value.Name.Should().Be("origin");
             RepositoryManager.ActiveRemote.Value.Url.Should().Be("https://github.com/EvilStanleyGoldman/IOTestsRepo.git");
 
+            RepositoryManager.Repository.CloneUrl.Should().Be("https://github.com/EvilStanleyGoldman/IOTestsRepo");
```
StanleyGoldman: @shana would you have expected this value?

src/tests/IntegrationTests/Events/RepositoryManagerTests.cs
```diff
@@ -305,6 +308,9 @@ await RepositoryManager
 
             RepositoryManager.ActiveRemote.HasValue.Should().BeFalse();
 
+            RepositoryManager.Repository.CloneUrl.Should().BeNull();
+            RepositoryManager.Repository.Owner.Should().BeNull();
```
StanleyGoldman: The cause of the bug is the reason why these tests fail

src/UnityExtension/Assets/Editor/GitHub.Unity/UI/HistoryView.cs
```diff
@@ -321,8 +323,14 @@ public void OnEmbeddedGUI()
 
                 GUILayout.FlexibleSpace();
 
+                GUI.enabled = currentRemote != null;
```
CapnRat: Should get into the habit of using [EditorGUI.Begin](https://docs.unity3d.com/ScriptReference/EditorGUI.BeginDisabledGroup.html)/[EndDisabledGroup](https://docs.unity3d.com/ScriptReference/EditorGUI.EndDisabledGroup.html) rather than setting GUI.enabled directly. Those utility methods handle nesting properly. (There's also [EditorGUI.DisabledScope](https://docs.unity3d.com/ScriptReference/EditorGUI.DisabledScope.html), for an IDisposable way of doing things.)

src/tests/IntegrationTests/Events/RepositoryManagerTests.cs
shana: Yeah? That looks about right

shana:
This PR needs more detail on exactly what is going wrong, with repro steps.

shana:
Actually, closing this as it has nothing other than tests, might as well fix the original branch.

Issue #23, opened 2017-06-27
Changes view performance problems


StanleyGoldman:
As described in https://github.com/github-beta/unity-preview/issues/29, the Changes view can't handle more than 1500 files (probably less on less performant machines), so we have some improvements to make on this area.

shiena:
Commit details view also has perfomance problems.
For example, this commit has changed 313 files. github-for-unity will be slower if you open it in the commit detail view.
https://github.com/unity3d-jp/piranhan/commit/7bbfaa9027456661c5a069e04b92ef2ff582e8af

meaghanlewis:
@StanleyGoldman this is an issue we can close out when we release the latest version, right?

StanleyGoldman:
![image](https://user-images.githubusercontent.com/417571/33799624-77dfce48-dcfd-11e7-841c-b54b9420ea25.png)

StanleyGoldman:
![](https://i.imgur.com/DW2n6S2.gif)

meaghanlewis:
this looks great 🎉 🎉 🎉

StanleyGoldman:
The `0.26` release solves these issues

PR #121, opened 2017-07-11
Refactoring and testing changes view tree model objects

StanleyGoldman:
In order to address #23 I am moving over the functionality to create the tree for the changes tree view

Depends on:
- [x] #120 
- [x] #133


shana:
FYI, PR #151 introduces a tree view control that would hopefully be useful for both branches and changes view. This PR looks to only move code around, right? I'm fine with merging this right now, then sit down and go through the tree view stuff and see if we can share it and fix these performance issues.

StanleyGoldman:
This PR only moves code into a place where I can test the output.

PR #123, opened 2017-07-12
correct win localappdata

thorade:
so that it can be copied&pasted


StanleyGoldman:
Hey @thorade I dont understand...

StanleyGoldman:
Thanks @thorade 👍

PR #119, opened 2017-07-11
Fix to improve performance of ChangesTreeView

StanleyGoldman:
Attempted #23 

performance test used pumps two sequences into `TreeBuilder.BuildTreeRoot()`:
1. Simple Test
  a. The first sequence contains 200 `GitStatusEntry` objects.
    1. There are 10 folders, with 10 files and 10 meta files.
  a. The second sequence contains 300 `GitStatusEntry` objects.
    1. Out of the original dataset every other file is retained.
      a. There are 10 folders, with 5 files and 5 meta files.
      a. Half of the original files can be considered "removed".
      a. An additional 10 folders, with 10 files and 10 meta files.
1. Advanced Test
  a. The first sequence contains 800 `GitStatusEntry` objects.
    1. There are 20 folders, with 20 files and 20 meta files.
  a. The second sequence contains 1200 `GitStatusEntry` objects.
    1. Out of the original dataset every other file is retained.
      a. There are 10 folders, with 10 files and 10 meta files.
      a. Half of the original files can be considered "removed".
      a. An additional 20 folders, with 20 files and 20 meta files.

|                       Method |     Mean |    Error |   StdDev |
|----------------------------- |---------|---------|---------|
 |Basic Test - Currently| 70.71 ms | 1.3993 ms | 1.3743 ms |
 |Basic Test - PR|   26.77 ms | 0.6483 ms | 0.6657 ms |
 |Heavy Test - Currently|754.46 ms | 6.6629 ms | 5.9064 ms |
 |Heavy Test - PR| 110.11 ms | 2.1563 ms | 2.2143 ms |

This branch will let you reproduce the performance test framework:
https://github.com/github-for-unity/Unity/tree/fixes/changes-view-performance-proof
Run the PerformanceTest project un-attached in Release mode.
It will take approx 8-9 minutes to complete.

Overall the PR starts at a third of the time of the current code base, and the amount of time it takes to execute grows bu half of the current code base.

Depends on:
- [x] #120 
- [x] #121
- [x] #133

PR #130, opened 2017-07-12
Fix to handle large git add operations

### LINKED ISSUE #105: Git Add fails due to too many files being added
### Steps to Reproduce

1. Open existing Unity project
2. Assets -> Import Package -> Import github-for-unity v0.15
3. create new repo from github web ui
4. in Unity, initialize and push (successfull)
5. got to changes, see 759 changes, select all, type commit message, commit

** Behavior:** 
Typing the commit message is very slow.
Hitting the commit button seems to result in no action.

**Reproduces how often:**
100% reproducible in this project

### Additional Information

Here is my log:

[github-unity_log.txt](https://github.com/github-for-unity/Unity/files/1135863/github-unity_log.txt)

StanleyGoldman:
Fixes #105 

Depends on:
- [x] #120 
- [x] #121
- [x] #133 
- [x] #119 
- [x] #140


src/GitHub.Api/Git/Tasks/GitAddTask.cs
```diff
@@ -22,6 +28,25 @@ class GitAddTask : ProcessTask<string>
...
+            arguments = "add ";
+
+            switch (addFileOption)
+            {
+                case AddFileOption.All:
+                    arguments += "-A";
+                    break;
+
+                case AddFileOption.CurrentDirectory:
+                    arguments += ".";
```
shana: Both these options do the same thing since we're always running git operations from the root of the repository with full paths to all files, so might as well not have them?

Issue #132, opened 2017-07-13
Changes View does not refresh on initialize


StanleyGoldman:
1. Start a new project with a new repository
2. Make some changes to the workspace
3. Close Unity while leaving the "ChangesView" selected/displayed
4. Open Unity and select project

Observe that there are changes in the workspace that should be displayed, but the Changes Tree View shows no changes. You have to click to the History tab and back before things refresh.

StanleyGoldman:
Fixed by #91

PR #134, opened 2017-07-13
Fixes to refresh ChangesView on start

StanleyGoldman:
Fixes: #132

### DIFF
--- a/src/UnityExtension/Assets/Editor/GitHub.Unity/UI/ChangesView.cs
+++ b/src/UnityExtension/Assets/Editor/GitHub.Unity/UI/ChangesView.cs
@@ -31,11 +31,7 @@ namespace GitHub.Unity
         {
             base.InitializeView(parent);
             tree.InitializeView(this);
-        }
 
-        public override void OnEnable()
-        {
-            base.OnEnable();
             if (Repository == null)
                 return;
 


src/UnityExtension/Assets/Editor/GitHub.Unity/UI/ChangesView.cs
```diff
@@ -31,11 +31,7 @@ public override void InitializeView(IView parent)
         {
             base.InitializeView(parent);
             tree.InitializeView(this);
-        }
 
-        public override void OnEnable()
-        {
-            base.OnEnable();
```
shana: Probably don't want to just to this always at initialize, since it calls Repository Refresh and all views get initialized. The logic of doing this on enable is correct, it's just that OnEnable might get called before the Repository instance is actually set, which means that we need to also do this whenever OnGUI is called. #91 introduces some methods that get called to update handlers and data, and the logic then becomes like this: https://github.com/github-for-unity/Unity/blob/3de69a70c0021b26e314b70116d111871e2e07ef/src/UnityExtension/Assets/Editor/GitHub.Unity/UI/BranchesView.cs#L50-L86

I could have sworn that that PR had already landed in master, really confused how it's not there yet, so that needs to land in order for these refresh bugs to get fixed.

StanleyGoldman:
Closing this out in favor of #91

PR #136, opened 2017-07-13
Fix garbled characters in history view with non-ascii characters (fixes#124)

### LINKED ISSUE #35: ProcessManagerIntegrationTests RussianLogEntriesTest no longer passes

### LINKED ISSUE #124: Non-ASCII character logs and file names are garbled in the history view and changes view.
history from https://github.com/unity3d-jp/piranhan

image

The reason is that `Process::OutputDataReceived` character-encodes with `Encoding.Default`, so the encoding output from the command does not match the encoding.

That is here.
https://github.com/Unity-Technologies/mono/blob/a449454f2128b2c0bb62d5ce3b8e22ab6b71db39/mcs/class/System/System.Diagnostics/Process.cs#L1424

**call hierarchy**
```
BeginOutputReadLine
-> ProcessAsyncReader
-> ProcessAsyncReader::AddInput
-> Encoding::Default::GetString
```

Since `Process::StandardOutput::ReadLine` uses `Process::StartInfo::StandardOutputEncoding`, combining `git -c i18n.logoutputencoding=utf8 log` and `Encoding.UTF8` will not garbled. However, since this is a synchronization process, Unity freezes until the command is finished.

Also, I have submitted a pull request on Unity's problem.
https://github.com/Unity-Technologies/mono/pull/582

For `Encoding.Default`,
* Japanese Windows codepage is 932
* OSX is the same as `Encoding.ASCII`

shiena:
fixes #124 
fixes #35 

Depends on:
- [x] #181

I replaced the reading of standard output from asynchronous to synchronous to eliminate garbled characters.

image

Synchronous reading of both standard output and standard error may cause deadlock. So I replaced standard output only.
Reference: https://stackoverflow.com/questions/7160187/standardoutput-readtoend-hangs

Also, I'm worried about whether performance will be bad.


shiena:
@StanleyGoldman 
The appveyor build failed with an error occurred in [`nuget restore GitHub.Unity.sln`](https://ci.appveyor.com/project/github-windows/unity/build/616.master#L268).
Can you build it again?

shiena:
@StanleyGoldman 
A pull request has been completed. I will summarize the changes.
* Using `Process::StandardOutput::BaseStream::BeginRead` instead of `Process::BeginOutputReadLine`
* Also change `StandardError`
* However, since .NET 4.6 has no problem in character encoding, use the original code
* Add options to git command according to [Git for Windows Unicode support](https://github.com/msysgit/msysgit/wiki/Git-for-Windows-Unicode-Support)

StanleyGoldman:
Hey @shiena sorry I've been MIA, @shana and I have been on vacation.
We will take a look at this soon.

StanleyGoldman:
Hey @shiena, sorry it's taken so long to get to this.

Processes are executed in a separate thread, so because of the bug in Unity we can definitely go with synchronous calls here. Also until Unity removes support for their .net 3.5 profile we will not compile this library in .net 4.6.

So when you get the chance remove all the `#if/#endif` and keep the .net 3.5 side. I think there are a few more changes to make so after that I will review again.

shiena:
@StanleyGoldman Thank you for review.
I kept the .net 3.5 side according to the compilation policy.

StanleyGoldman:
💨... Thanks for the quick changes ...
I'm testing your changes against the https://github.com/unity3d-jp/piranhan repo now.

src/GitHub.Api/NewTaskSystem/ProcessTask.cs
```diff
@@ -74,39 +78,74 @@ class ProcessWrapper
...
-                    //Logger.Trace("OutputData \"" + (e.Data == null ? "'null'" : e.Data) + "\"");
-
-                    string encodedData = null;
-                    if (e.Data != null)
+                    int bytesRead = Process.StandardOutput.BaseStream.EndRead(r);
+                    if (bytesRead > 0)
+                    {
+                        var encoded = outputEncoding.GetString(outputBuffer, 0, bytesRead);
+                        outputContents.Append(encoded);
+                        Process.StandardOutput.BaseStream.BeginRead(outputBuffer, 0, outputBuffer.Length, outputCallback, Process.StandardOutput);
```
StanleyGoldman: I believe this line of code should not be here.

src/GitHub.Api/NewTaskSystem/ProcessTask.cs
```diff
@@ -74,39 +78,74 @@ class ProcessWrapper
...
-                    //}
-
-                    string encodedData = null;
-                    if (e.Data != null)
+                    int bytesRead = Process.StandardError.BaseStream.EndRead(r);
+                    if (bytesRead > 0)
+                    {
+                        var encoded = errorEncoding.GetString(errorBuffer, 0, bytesRead);
+                        errorContents.Append(encoded);
+                        Process.StandardError.BaseStream.BeginRead(errorBuffer, 0, errorBuffer.Length, errorCallback, Process.StandardError);
```
StanleyGoldman: I think the same goes for this line.

src/GitHub.Api/NewTaskSystem/ProcessTask.cs
StanleyGoldman: It seems I am mistaken

src/GitHub.Api/NewTaskSystem/ProcessTask.cs
StanleyGoldman: Mistaken here too

StanleyGoldman:
I understand why you've done what you've done... 🤔... I'm thinking about it

src/GitHub.Api/NewTaskSystem/ProcessTask.cs
```diff
@@ -134,14 +173,37 @@ public void Run()
             }
 
             if (Process.StartInfo.RedirectStandardOutput)
-                Process.BeginOutputReadLine();
+            {
+                outputReset.Reset();
+                Process.StandardOutput.BaseStream.BeginRead(
```
StanleyGoldman: Use a loop here to read the bytes and encode the string with the correct encoding. After that you can read the lines and call the appropriate output method. This way we won't need the complicated `AsyncCallback` that has a reference to itself.

src/GitHub.Api/NewTaskSystem/ProcessTask.cs
```diff
@@ -134,14 +173,37 @@ public void Run()
...
+                    outputBuffer.Length,
+                    outputCallback,
+                    Process.StandardOutput
+                    );
+            }
             if (Process.StartInfo.RedirectStandardError)
-                Process.BeginErrorReadLine();
+            {
+                errorReset.Reset();
+                Process.StandardError.BaseStream.BeginRead(
```
StanleyGoldman: Do the same thing here.

src/GitHub.Api/NewTaskSystem/ProcessTask.cs
```diff
@@ -134,14 +173,37 @@ public void Run()
...
+            }
             if (Process.StartInfo.RedirectStandardInput)
                 Input = new StreamWriter(Process.StandardInput.BaseStream, new UTF8Encoding(false));
 
             onStart?.Invoke();
 
+            if (Process.StartInfo.RedirectStandardOutput)
+                outputReset.WaitOne();
+            if (Process.StartInfo.RedirectStandardError)
+                errorReset.WaitOne();
```
StanleyGoldman: That way we no longer need these `ManualResetEvent` objects.

src/GitHub.Api/NewTaskSystem/ProcessTask.cs
shiena: If you read the standard output using a loop there, it blocks processing.
Because `RunCredentialHelper` writes to the standard input of `git-credential-wincred` command when signing in to GitHub.
https://github.com/github-for-unity/Unity/blob/1476044c625f615e74b282ecb5c4c286fe0bb86b/src/GitHub.Api/Git/GitCredentialManager.cs#L147-L172

For that reason, I think that we need an asynchronous `BeginRead`.
Please tell me if there is a better way.

StanleyGoldman:
Hey @shiena I'd like to add some commits to this PR. There is a checkbox somewhere in this PR to let the project maintainer add commits to this PR if you edit it. I have a fix I'd like you to try.

StanleyGoldman:
Hey @shiena even though I'm supposed to be able to push to the branch, I'm not able to.

Could you merge [this branch](https://github.com/github-for-unity/Unity/tree/fix-history-view-encoding) into your PR and test it out.

StanleyGoldman:
I get the impression I need to set Japanese as my default language in Windows to properly debug this error.

shiena:
@StanleyGoldman I merge and tested [your branch](https://github.com/github-for-unity/Unity/tree/fix-history-view-encoding).
Japanese log was correctly displayed.

StanleyGoldman:
@shiena there is another branch based off this, could you test it out for me.

https://github.com/github-for-unity/Unity/tree/fixes/returning-lines-sooner

StanleyGoldman:
LFS Verification checks are preventing me from pushing to your repo

shiena:
@StanleyGoldman Japanese also displayed correctly for that branch.

StanleyGoldman:
Edited the pull request description to show what this PR depends on

shiena:
@StanleyGoldman I think that https://github.com/github-for-unity/Unity/issues/35 has also been fixed.

StanleyGoldman:
You may be right, let's see

StanleyGoldman:
You are correct.

src/GitHub.Api/NewTaskSystem/ProcessTask.cs
```diff
@@ -128,54 +96,103 @@ public void Run()
...
-                while (!WaitForExit(500))
+                // buffer size refers to https://github.com/Unity-Technologies/mono/blob/unity-5.6-staging/mcs/class/System/System.Diagnostics/Process.cs#L1149-L1157
+                const int bufferSize = 8182;
+
+                if (Process.StartInfo.RedirectStandardOutput)
                 {
-                    if (token.IsCancellationRequested)
+                    var outputStream = Process.StandardOutput.BaseStream;
+                    var outputBuffer = new byte[bufferSize];
+                    var outputEncoding = Process.StartInfo.StandardOutputEncoding ?? Console.Out.Encoding;
```
StanleyGoldman: Can `Process.StartInfo.StandardOutputEncoding` ever be null?

src/GitHub.Api/NewTaskSystem/ProcessTask.cs
```diff
@@ -128,54 +96,103 @@ public void Run()
...
+                if (Process.StartInfo.RedirectStandardError)
                 {
-                    onError?.Invoke(null, String.Join(Environment.NewLine, errors.ToArray()));
-                }
-            }
-            onEnd?.Invoke();
-        }
+                    var errorStream = Process.StandardError.BaseStream;
+                    var errorBuffer = new byte[bufferSize];
+                    var errorEncoding = Process.StartInfo.StandardErrorEncoding ?? Console.Error.Encoding;
```
StanleyGoldman: Same question about `Process.StartInfo.StandardErrorEncoding` ...

src/GitHub.Api/NewTaskSystem/ProcessTask.cs
shiena: It could not be null by tracing the call hierarchy. So delete the null check.

src/GitHub.Api/NewTaskSystem/ProcessTask.cs
shiena: Also delete null check.

shiena:
@StanleyGoldman @shana 
A process encoding bug has been avoided as https://github.com/github-for-unity/Unity/pull/181 has been merged. (`Process.StandardOutput.ReadLine` correctly encodes)
However, if the user sets encoding options other than `utf8` in git config, the history view garbled.
Therefore, I recommend setting `-C logoutputencoding=utf8` etc for git command line option.

Even if you need the git command line option, this pull request is redundant, so I would like to create another pull request.

shana:
@shiena Yup, we were literally just discussing that 10 minutes ago and came to the same conclusion 😄 👍

StanleyGoldman:
Hey @shiena we were about to say the same thing to you...

PR #141, opened 2017-07-14
Repository Refactor

StanleyGoldman:
I'm trying to clean up the `Repository` class.
It has a `IGitClient` that it doesn't need, I removed that from the constructor.

I also changed `CloneUrl` to be driven from a field in the `RepositoryManager`.
Currently, the value is figured out in the `RepositoryManager` and set in the constructor of the `Repository`.
On a config change event the `RepositoryManager` raises `OnRemoteOrTrackingChanged`, and the `Repository` has a duplicate of the logic to set it's own `CloneUrl`.

Depends on:
- [x] #139

### DIFF
--- a/src/GitHub.Api/Git/GitConfig.cs
+++ b/src/GitHub.Api/Git/GitConfig.cs
@@ -1,6 +1,5 @@
 using System;
 using System.Collections.Generic;
-using System.IO;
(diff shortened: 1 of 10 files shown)


StanleyGoldman:
Closing due to all the action in #91

PR #149, opened 2017-07-28
Support paths with spaces in the package.sh script

shana:
This is what I get for installing Unity in paths without spaces in them...

This PR also changes the path you need to give to package.sh to build the package on Windows, so it matches the logic on the mac


shana:
Superseded by #129

PR #151, opened 2017-07-31
Tree view control for displaying data performantly.

shana:
**Note:** This branch targets `enhancements/branches-view-rollup` #464 

This PR introduces a tree view control to display data in a tree view in a performant way, since the layouting system can't really handle a lot of the data we throw into it.

The approach here is to separate the data from the UI - the data gets put into a data structure (with whatever necessary processing), and then passed into the UI structure that will render it. Both can be cached separately and independently.

This tries to solve a number of problems:

- Unity can't serialize data recursively (like a tree of nodes that have child nodes), so the data that is fed to the UI must be serialized in a flat structure
- The layout system has problems when nodes are collapsed, changing the padding and spacing of items
- We need to cache the UI so that it doesn't disappear when Unity reloads the domain. We also need to cache the data that feeds the UI so we can show existing data while we're refreshing it.

Missing in this PR

- Scrolling support, rendering only the area that needs to be displayed, so that we don't try to render a 1000 objects

This is currently applied to the branches view but the Treeview control should be generic enough to be applied to the changes view as well, hopefully fixing the performance problems there.


StanleyGoldman:
Testing the latest, it is looking good. Noticed this display issue with remotes and multiple remotes at that.

img

StanleyGoldman:
I sent this PR a PR.. #372

StanleyGoldman:
Sending you a pull request in #378

StanleyGoldman:
Making an edit to say this depends on #384

src/UnityExtension/Assets/Editor/GitHub.Unity/UI/TreeControl.cs
```diff
@@ -0,0 +1,465 @@
...
+
+        [SerializeField] private SerializableTexture2D folderIcon = new SerializableTexture2D();
+        public Texture2D FolderIcon {  get { return folderIcon.Texture; } set { folderIcon.Texture = value; } }
+
+        [SerializeField] private SerializableTexture2D rootFolderIcon = new SerializableTexture2D();
+        public Texture2D RootFolderIcon {  get { return rootFolderIcon.Texture; } set { rootFolderIcon.Texture = value; } }
+
+        [SerializeField] public GUIStyle FolderStyle;
+        [SerializeField] public GUIStyle TreeNodeStyle;
+        [SerializeField] public GUIStyle ActiveTreeNodeStyle;
```
CapnRat: Normally I wouldn't serialize GUIStyles/Textures, instead I load them on demand (typically in OnEnable). This makes it more difficult to iterate on the source styles/textures without recreating the serialized object completely. Nothing particularly wrong with this approach though.

src/UnityExtension/Assets/Editor/GitHub.Unity/UI/TreeControl.cs
```diff
@@ -0,0 +1,465 @@
...
+                style = IsActive ? activeNodeStyle : nodeStyle;
+            }
+
+            bool changed = false;
+            var fillRect = rect;
+            var nodeRect = new Rect(Level * indentation, rect.y, rect.width, rect.height);
+
+            if (Event.current.type == EventType.repaint)
+            {
+                nodeStyle.Draw(fillRect, "", false, false, false, isSelected);
```
CapnRat: GUIContent.none rather than empty string.

src/UnityExtension/Assets/Editor/GitHub.Unity/UI/TreeControl.cs
```diff
@@ -0,0 +1,465 @@
...
+using UnityEngine.Profiling;
+
+namespace GitHub.Unity
+{
+    [Serializable]
+    public class Tree
+    {
+        [SerializeField] public float ItemHeight = EditorGUIUtility.singleLineHeight;
+        [SerializeField] public float ItemSpacing = EditorGUIUtility.standardVerticalSpacing;
+        [SerializeField] public float Indentation = 12f;
```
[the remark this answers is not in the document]
StanleyGoldman: 👍

src/UnityExtension/Assets/Editor/GitHub.Unity/UI/TreeControl.cs
```diff
@@ -0,0 +1,465 @@
...
+                SelectedNode = rootNode;
+            }
+            return idx;
+        }
+
+        private bool HandleInput(Rect rect, TreeNode currentNode, int index, Action<TreeNode> singleClick = null, Action<TreeNode> doubleClick = null)
+        {
+            bool selectionChanged = false;
+            var clickRect = new Rect(0f, rect.y, rect.width, rect.height);
+            if (Event.current.type == EventType.MouseDown && clickRect.Contains(Event.current.mousePosition))
```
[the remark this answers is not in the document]
StanleyGoldman: That is correct. The fix to this is coming in another pull request.

src/UnityExtension/Assets/Editor/GitHub.Unity/UI/TreeControl.cs
StanleyGoldman: 👍

src/UnityExtension/Assets/Editor/GitHub.Unity/UI/TreeControl.cs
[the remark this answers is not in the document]
StanleyGoldman: #467

src/UnityExtension/Assets/Editor/GitHub.Unity/UI/TreeControl.cs
```diff
@@ -0,0 +1,465 @@
...
+            titleNode.Load();
+            nodes.Add(titleNode);
+
+            foreach (var d in data)
+            {
+                var parts = d.Name.Split('/');
+                for (int i = 0; i < parts.Length; i++)
+                {
+                    var label = parts[i];
+                    var name = String.Join("/", parts, 0, i + 1);
```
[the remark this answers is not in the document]
StanleyGoldman: I'm about to add functionality to the TreeControl to display a different type of tree. So these lines of codes are going to be a stomping ground for a little bit. I'm going to log an issue under `tech-debt` for this.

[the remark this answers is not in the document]
StanleyGoldman: #477

src/UnityExtension/Assets/Editor/GitHub.Unity/UI/BranchesView.cs
```diff
@@ -33,42 +33,34 @@ class BranchesView : Subview
...
         public override void OnEnable()
         {
             base.OnEnable();
             AttachHandlers(Repository);
-
-            if (Repository != null)
-            {
-                Repository.CheckLocalAndRemoteBranchListChangedEvent(lastLocalAndRemoteBranchListChangedEvent);
-            }
+            Repository.CheckLocalAndRemoteBranchListChangedEvent(lastLocalAndRemoteBranchListChangedEvent);
```
sguthals: Repository is not longer possibly null then?

src/UnityExtension/Assets/Editor/GitHub.Unity/UI/BranchesView.cs
```diff
@@ -536,95 +348,81 @@ private void OnTreeNodeGUI(BranchTreeNode node)
...
+
+            //Debug.LogFormat("reserving: {0} {1} {2}", rect.y - initialRect.y, rect.y, initialRect.y);
+            GUILayout.Space(rect.y - initialRect.y);
         }
 
-        private void OnTreeNodeChildrenGUI(BranchTreeNode node)
+        private int CompareBranches(GitBranch a, GitBranch b)
         {
-            if (node == null || node.Children == null)
+            //if (IsFavorite(a.Name))
```
sguthals: meant to be commented out? Should it just be removed?

src/UnityExtension/Assets/Editor/GitHub.Unity/UI/BranchesView.cs
```diff
@@ -536,95 +348,81 @@ private void OnTreeNodeGUI(BranchTreeNode node)
...
-                        Event.current.Use();
-                    }
-                }
+                return 1;
             }
+
+            return a.Name.CompareTo(b.Name);
         }
 
+        //private bool IsFavorite(string branchName)
```
sguthals: Commented or deleted?

src/UnityExtension/Assets/Editor/GitHub.Unity/UI/TreeControl.cs
```diff
@@ -0,0 +1,465 @@
...
+        public void Focus()
+        {
+            bool selectionChanged = false;
+            if (Event.current.type == EventType.KeyDown)
+            {
+                int directionY = Event.current.keyCode == KeyCode.UpArrow ? -1 : Event.current.keyCode == KeyCode.DownArrow ? 1 : 0;
+                int directionX = Event.current.keyCode == KeyCode.LeftArrow ? -1 : Event.current.keyCode == KeyCode.RightArrow ? 1 : 0;
+                if (directionY != 0 || directionX != 0)
+                {
+                    if (directionY < 0 || directionY < 0)
```
sguthals: Should be directionX on the second condition

src/UnityExtension/Assets/Editor/GitHub.Unity/UI/TreeControl.cs
```diff
@@ -0,0 +1,465 @@
...
+        }
+
+        public void Focus()
+        {
+            bool selectionChanged = false;
+            if (Event.current.type == EventType.KeyDown)
+            {
+                int directionY = Event.current.keyCode == KeyCode.UpArrow ? -1 : Event.current.keyCode == KeyCode.DownArrow ? 1 : 0;
+                int directionX = Event.current.keyCode == KeyCode.LeftArrow ? -1 : Event.current.keyCode == KeyCode.RightArrow ? 1 : 0;
+                if (directionY != 0 || directionX != 0)
```
sguthals: I'm a bit confused by this set of if statements. I don't fully understand the functionality here, so that just might be my confusion, but it seems like a set of nested ifs with two different variables shouldn't just be an or?

src/UnityExtension/Assets/Editor/GitHub.Unity/UI/TreeControl.cs
```diff
@@ -0,0 +1,465 @@
...
+                }
+            }
+            if (SelectedNode != null && SelectedNode.IsHidden)
+            {
+                SelectedNode = rootNode;
+            }
+            return idx;
+        }
+
+        private bool HandleInput(Rect rect, TreeNode currentNode, int index, Action<TreeNode> singleClick = null, Action<TreeNode> doubleClick = null)
```
sguthals: So are singleClick and doubleClick what gets displayed if a node gets single or double clicked?

src/UnityExtension/Assets/Editor/GitHub.Unity/UI/BranchesView.cs
StanleyGoldman: By the time the `BranchesView` is displayed we always have a `Repository`

src/UnityExtension/Assets/Editor/GitHub.Unity/UI/BranchesView.cs
StanleyGoldman: That is getting cleaned up in #468

src/UnityExtension/Assets/Editor/GitHub.Unity/UI/BranchesView.cs
StanleyGoldman: That is getting cleaned up in #468

src/UnityExtension/Assets/Editor/GitHub.Unity/UI/TreeControl.cs
StanleyGoldman: Almost... they are the function delegates that get called when a node gets single or double clicked

src/UnityExtension/Assets/Editor/GitHub.Unity/UI/TreeControl.cs
StanleyGoldman: You are right, the wrapping if statement is redundant.

The local and remote trees are two separate tree controls.
This code lets you travel from one tree control to the other via keystroke if the scenario is correct.

src/UnityExtension/Assets/Editor/GitHub.Unity/UI/TreeControl.cs
StanleyGoldman: Good catch!

src/UnityExtension/Assets/Editor/GitHub.Unity/UI/TreeControl.cs
sguthals: Got it, thanks!

PR #187, opened 2017-08-14
Fix to not load the Keychain on application startup

### LINKED ISSUE #60: Fatal Error "CheckDisalowAllocation" [sic] when opening GitHub Window
Imported v0.12 alpha into a project in Unity 5.6.1p4 on Mac OS X 10.12.5, then selected GitHub/Window from the main menu. This popped up:

image

Then the entire editor became unresponsive and had to be force closed.

Correction: The ~/.local/share/GitHubUnity/github-unity.log was not generated at all - there's still an old one on my system from earlier versions of this plugin.

StanleyGoldman:
Fixes #60


shana:
Unity enters playmode and restarts everything, we do `RunCredentialHelper helper:"osxkeychain" action:"get"` and we write to the input stream the user we’re trying to get, and while we’re doing this the user stops Unity, which triggers a GC collection, which goes up the tree looking for objects to collect and probably crashes on the stream access that we’re currently doing.

PR #193, opened 2017-08-15
Add symlink resolver so we can figure out paths properly

### LINKED ISSUE #192: Push/pull/publish randomly failing on non-windows systems
Git apparently can't find helper tools like `git-remote-https` on mac and linux, causing network operations to fail (and possibly others as well).

This is caused by the path to git being a symlink to a different directory, causing us to set the wrong paths in the environment.

shana:
Fixes #192

On non-windows, git and other files might be symlinks, so when we're using them as a base for determining parent directories, we may need to resolve the symlink first. The rules for automatically
doing symlink resolution are tricky, so for now just do it manually.

Still not happy with how we're finding git on the system, but this should at least make the command line work across all systems. 

~~An existing problem right now is that OSX overwrites values set to the PATH env variable when it loads `/etc/profile` in preparation to run bash, which is seriously screwed up on their part. That means that the opened command line won't have our path variable value (but it will have all the other env vars). I'm still investigating how to fix that...~~ This is fixed now by creating a little shell script that in turn invokes bash with a proper environment.

FYI, we want the GitExecutablePath to reflect where the binary was found, and the GitInstallPath to reflect where it is installed. When the binary is a symlink (like /usr/local/bin/git -> ../Cellar/git/2.12.2/bin/git), this means GitExecutablePath will have /usr/local/bin/git and GitInstallPath will have /usr/local/Cellar/git/2.12.2/bin

### DIFF
--- a/src/GitHub.Api/IO/NiceIO.cs
+++ b/src/GitHub.Api/IO/NiceIO.cs
@@ -283,6 +283,8 @@ GitHub.Unity
 
         public bool DirectoryExists(NPath append)
         {
+            if (append == null)
+                return FileSystem.DirectoryExists(ToString());
             return FileSystem.DirectoryExists(Combine(append).ToString());
         }
 
@@ -295,6 +297,8 @@ GitHub.Unity
 
         public bool FileExists(NPath append)
         {
+            if (append == null)
+                return FileSystem.FileExists(ToString());
             return FileSystem.FileExists(Combine(append).ToString());
         }
 
@@ -1017,6 +1021,14 @@ GitHub.Unity
             if (path == null)
                 return null;
             return new NPath(path);
+        }
+
+        public static NPath Resolve(this NPath path)
+        {
+            if (path == null || DefaultEnvironment.OnWindows || path.IsRelative || !path.FileExists())
+                return path;
+
+            return new NPath(Mono.Unix.UnixPath.GetCompleteRealPath(path.ToString()));
         }
     }
 
--- a/src/GitHub.Api/IO/NiceIO.cs
+++ b/src/GitHub.Api/IO/NiceIO.cs
@@ -753,11 +753,14 @@ GitHub.Unity
             }
         }
 
+        private static NPath systemTemp;
         public static NPath SystemTemp
         {
             get
             {
-                return new NPath(FileSystem.GetTempPath());
+                if (systemTemp == null)
+                    systemTemp = new NPath(FileSystem.GetTempPath());
+                return systemTemp;
             }
         }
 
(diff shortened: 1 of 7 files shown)


src/GitHub.Api/IO/NiceIO.cs
```diff
@@ -283,6 +283,8 @@ public bool DirectoryExists(string append = "")
 
         public bool DirectoryExists(NPath append)
         {
+            if (append == null)
```
StanleyGoldman: If this method was sent null before, what was the result?

src/GitHub.Api/IO/NiceIO.cs
```diff
@@ -1018,6 +1025,14 @@ public static NPath ToNPath(this string path)
                 return null;
             return new NPath(path);
         }
+
+        public static NPath Resolve(this NPath path)
+        {
+            if (path == null || DefaultEnvironment.OnWindows || path.IsRelative || !path.FileExists())
+                return path;
+
+            return new NPath(Mono.Unix.UnixPath.GetCompleteRealPath(path.ToString()));
```
StanleyGoldman: `GetCompleteRealPath` is a method that will get the real link from a symlink?

src/GitHub.Api/OutputProcessors/ProcessManager.cs
```diff
@@ -74,8 +74,13 @@ public void RunCommandLineWindow(NPath workingDirectory)
             }
             else if (environment.IsMac)
             {
+                // we need to create a temp bash script to set up the environment properly, because
+                // osx terminal app doesn't inherit the PATH env var and there's no way to pass it in
+                var envVarFile = environment.FileSystem.GetRandomFileName();
+                environment.FileSystem.WriteAllLines(envVarFile, new string[] { "cd $GHU_WORKINGDIR", "PATH=$GHU_FULLPATH:$PATH /bin/bash" });
+                Mono.Unix.Native.Syscall.chmod(envVarFile, (Mono.Unix.Native.FilePermissions)493); // -rwxr-xr-x mode (0755)
```
StanleyGoldman: 🎖

src/GitHub.Api/IO/NiceIO.cs
shana: Yeah, from Mono.Posix

src/GitHub.Api/OutputProcessors/ProcessManager.cs
shana: magic

src/GitHub.Api/IO/NiceIO.cs
shana: `Combine` doesn't care if it's null so it wouldn't blow up, but it's a waste of resources creating new instances of the same thing over and over again, `Combine` always creates a new thing.

StanleyGoldman: Ah, I follow now

PR #195, opened 2017-08-15
Fixes to populate remote url text field after publish

### LINKED ISSUE #185: Remote URL blank after publish
tried this 6 times already, with 2 different copies of unity, but upon start lowest status bar consists of "initializing github for unity log file [...]" for hours. I decided to ignore it and proceeded to log in and link up my (currently empty) project to gitub repo, after pasting link to .git and clicking "save repository" plugin's settings UI becomes inactive, then entire unity slows down witin minutes. No matter what I change in my config, this is 6th time it does exactly that.

StanleyGoldman:
Fixes #185 

The problem here was due to the use of `activeRemote` in `SettingsView`.

When a user is publishing, they are looking at the `HistoryView`. Which means the `SettingsView` is disabled, and when a remote is added the event handler `SettingsView.OnRepositoryChanged` will not run and won't update `activeRemote` with the new value.

I removed `activeRemote` and switched it to `Repository.CurrentRemote`.

I also fixed two errors in the `SettingsView` that would appear in the user log.

### DIFF
--- a/src/UnityExtension/Assets/Editor/GitHub.Unity/UI/SettingsView.cs
+++ b/src/UnityExtension/Assets/Editor/GitHub.Unity/UI/SettingsView.cs
@@ -67,7 +67,6 @@ namespace GitHub.Unity
         private const string DefaultRepositoryRemoteName = "origin";
 
         [NonSerialized] private int newGitIgnoreRulesSelection = -1;
-        [NonSerialized] private ConfigRemote? activeRemote;
 
         [SerializeField] private string gitName;
         [SerializeField] private string gitEmail;
@@ -94,6 +93,8 @@ namespace GitHub.Unity
         {
             base.OnEnable();
             AttachHandlers(Repository);
+
+            remoteHasChanged = true;
         }
 
         public override void OnDisable()
@@ -114,8 +115,9 @@ namespace GitHub.Unity
 
             DetachHandlers(oldRepository);
             AttachHandlers(Repository);
-            activeRemote = Repository.CurrentRemote;
+
             remoteHasChanged = true;
+
             Refresh();
         }
 
@@ -178,7 +180,7 @@ namespace GitHub.Unity
 
             if (Repository == null)
             {
-                if (cachedUser == null || String.IsNullOrEmpty(cachedUser.Name))
+                if ((cachedUser == null || String.IsNullOrEmpty(cachedUser.Name)) && GitClient != null)
                 {
                     var user = new User();
                     GitClient.GetConfig("user.name", GitConfigSource.User)
@@ -221,6 +223,7 @@ namespace GitHub.Unity
             if (remoteHasChanged)
             {
                 remoteHasChanged = false;
+                var activeRemote = Repository.CurrentRemote;
                 hasRemote = activeRemote.HasValue && !String.IsNullOrEmpty(activeRemote.Value.Url);
                 if (!hasRemote)
                 {
@@ -245,7 +248,6 @@ namespace GitHub.Unity
 
         private void Repository_OnActiveRemoteChanged(string remote)
         {
-            activeRemote = Repository.CurrentRemote;
             remoteHasChanged = true;
         }
 


src/UnityExtension/Assets/Editor/GitHub.Unity/UI/SettingsView.cs
```diff
@@ -180,21 +180,24 @@ private void MaybeUpdateData()
...
                 {
-                    var user = new User();
-                    GitClient.GetConfig("user.name", GitConfigSource.User)
-                        .Then((success, value) => user.Name = value).Then(
-                    GitClient.GetConfig("user.email", GitConfigSource.User)
-                        .Then((success, value) => user.Email = value))
-                    .FinallyInUI((success, ex) =>
-                    {
-                        if (success && !String.IsNullOrEmpty(user.Name))
+                    if(GitClient != null)
```
shana: hmmm why is `GitClient` null here? is the view racing the EntryPoint initialize code?
Also, spaaaaaaaaace.

src/UnityExtension/Assets/Editor/GitHub.Unity/UI/SettingsView.cs
```diff
@@ -240,12 +243,11 @@ private void ResetToDefaults()
             gitName = Repository != null ? Repository.User.Name : String.Empty;
             gitEmail = Repository != null ? Repository.User.Email : String.Empty;
             repositoryRemoteName = DefaultRepositoryRemoteName;
-            repositoryRemoteUrl = string.Empty;
+            newRepositoryRemoteUrl = repositoryRemoteUrl = string.Empty;
```
shana: This isn't being called anywhere, let's nuke it

src/UnityExtension/Assets/Editor/GitHub.Unity/UI/SettingsView.cs
shana: This if should be with the condition above:

`if ((cachedUser == null || String.IsNullOrEmpty(cachedUser.Name)) && GitClient != null)`

src/UnityExtension/Assets/Editor/GitHub.Unity/UI/SettingsView.cs
StanleyGoldman: `newRepositoryRemoteUrl` and `repositoryRemoteUrl` are both used..

https://github.com/github-for-unity/Unity/blob/4a208bd88f1396e135a89088dd1f25d69b96b653/src/UnityExtension/Assets/Editor/GitHub.Unity/UI/SettingsView.cs#L344-L345

src/UnityExtension/Assets/Editor/GitHub.Unity/UI/SettingsView.cs
StanleyGoldman: `GitClient` was null for me when I rebuilt code with Unity running. I then figured it was possible during a domain reload.

StanleyGoldman: I don't know which space we are referring to. 🙈

src/UnityExtension/Assets/Editor/GitHub.Unity/UI/SettingsView.cs
```diff
@@ -94,6 +94,9 @@ public override void OnEnable()
         {
             base.OnEnable();
             AttachHandlers(Repository);
+
+            activeRemote = Repository != null ? Repository.CurrentRemote : null;
+            remoteHasChanged = true;
```
StanleyGoldman: Setting `remoteHasChanged` is not enough. We need to update `activeRemote` at some point.
Part of my thinking was that `activeRemote` was only read from here..

https://github.com/github-for-unity/Unity/blob/30e665cad007f96223c247237d9f2de830bb60cd/src/UnityExtension/Assets/Editor/GitHub.Unity/UI/SettingsView.cs#L229-L239

shana: Yeah, you're right, there's no point in having the field if it's only ever touched here, and it shouldn't be touched anywhere else anyway

PR #197, opened 2017-08-16
Fix to disable PublishView until ready

### LINKED ISSUE #196: Publish dialog should be disabled while User/Organizations are loading

StanleyGoldman:
Fixes #196 

Made use of `BeginDisabledGroup`/`EndDisabledGroup`.

### DIFF
--- a/src/GitHub.Api/Application/ApiClient.cs
+++ b/src/GitHub.Api/Application/ApiClient.cs
@@ -190,7 +190,7 @@ namespace GitHub.Unity
 
                 if (!await EnsureKeychainLoaded())
                 {
-                    return null;
+                    throw new InvalidOperationException("The keychain did not load");
                 }
 
                 Octokit.Repository repository;
--- a/src/GitHub.Api/Application/ApiClient.cs
+++ b/src/GitHub.Api/Application/ApiClient.cs
@@ -61,7 +61,15 @@ namespace GitHub.Unity
         public async Task CreateRepository(NewRepository newRepository, Action<Octokit.Repository, Exception> callback, string organization = null)
         {
             Guard.ArgumentNotNull(callback, "callback");
-            await CreateRepositoryInternal(newRepository, callback, organization);
+            try
+            {
+                var repository = await CreateRepositoryInternal(newRepository, organization);
+                callback(repository, null);
+            }
+            catch (Exception e)
+            {
+                callback(null, e);
+            }
         }
 
         public async Task GetOrganizations(Action<IList<Organization>> callback)
@@ -174,7 +182,7 @@ namespace GitHub.Unity
             return result.Code == LoginResultCodes.Success;
         }
 
-        private async Task CreateRepositoryInternal(NewRepository newRepository, Action<Octokit.Repository, Exception> callback, string organization)
+        private async Task<Octokit.Repository> CreateRepositoryInternal(NewRepository newRepository, string organization)
         {
             try
             {
@@ -182,8 +190,7 @@ namespace GitHub.Unity
 
                 if (!await EnsureKeychainLoaded())
                 {
-                    callback(null, new Exception("Keychain Not Loaded"));
-                    return;
+                    return null;
                 }
 
                 Octokit.Repository repository;
@@ -201,13 +208,12 @@ namespace GitHub.Unity
                 }
 
                 logger.Trace("Created Repository");
-
-                callback(repository, null);
+                return repository;
             }
             catch (Exception ex)
             {
                 logger.Error(ex, "Error Creating Repository");
-                callback(null, ex);
+                throw;
             }
         }
 
--- a/src/UnityExtension/Assets/Editor/GitHub.Unity/UI/PUblishView.cs
+++ b/src/UnityExtension/Assets/Editor/GitHub.Unity/UI/PUblishView.cs
@@ -13,6 +13,10 @@ namespace GitHub.Unity
         private const string PrivateRepoMessage = "You choose who can see and commit to this repository";
         private const string PublicRepoMessage = "Anyone can see this repository. You choose who can commit";
         private const string PublishViewCreateButton = "Create";
+        private const string SelectedOwnerLabel = "Owner";
+        private const string RepositoryNameLabel = "Repository Name";
+        private const string DescriptionLabel = "Description";
+        private const string CreatePrivateRepositoryLabel = "Create as a private repository";
 
         [SerializeField] private string username;
         [SerializeField] private string[] owners = { };
@@ -138,7 +142,7 @@ namespace GitHub.Unity
                 {
                     GUILayout.BeginVertical();
                     {
-                        GUILayout.Label("Owner");
+                        GUILayout.Label(SelectedOwnerLabel);
 
                         selectedOwner = EditorGUILayout.Popup(0, owners);
                     }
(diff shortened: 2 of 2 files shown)


src/UnityExtension/Assets/Editor/GitHub.Unity/UI/PUblishView.cs
```diff
@@ -63,6 +65,8 @@ private void PopulateView()
                 {
                     Logger.Trace("GetCurrentUser");
 
+                    isLoading = true;
```
shana: Why not use `isBusy` for this?

src/UnityExtension/Assets/Editor/GitHub.Unity/UI/PUblishView.cs
```diff
@@ -127,111 +133,111 @@ public override void OnGUI()
...
+                GUILayout.BeginVertical();
                 {
-                    GUI.enabled = !isBusy;
-                    togglePrivate = GUILayout.Toggle(togglePrivate, "Create as a private repository");
-                    GUI.enabled = true;
-                }
-                GUILayout.EndHorizontal();
+                    GUILayout.BeginHorizontal();
+                    {
+                        togglePrivate = GUILayout.Toggle(togglePrivate, "Create as a private repository");
```
shana: Move this string to a const while you're at it

src/UnityExtension/Assets/Editor/GitHub.Unity/UI/PUblishView.cs
```diff
@@ -127,111 +133,111 @@ public override void OnGUI()
...
-            }
-            GUILayout.EndHorizontal();
+                GUILayout.EndHorizontal();
 
-            GUILayout.Label("Description");
-            GUI.enabled = !isBusy;
-            repoDescription = EditorGUILayout.TextField(repoDescription);
-            GUI.enabled = true;
-            GUILayout.Space(Styles.PublishViewSpacingHeight);
+                GUILayout.Label("Description");
```
shana: Move this string to a const while you're at it

src/UnityExtension/Assets/Editor/GitHub.Unity/UI/PUblishView.cs
```diff
@@ -127,111 +133,111 @@ public override void OnGUI()
...
 
-                GUILayout.BeginVertical();
-                {
-                    GUILayout.Label("Repository Name");
-                    GUI.enabled = !isBusy;
-                    repoName = EditorGUILayout.TextField(repoName);
-                    GUI.enabled = true;
+                    GUILayout.BeginVertical();
+                    {
+                        GUILayout.Label("Repository Name");
```
shana: Move this string to a const while you're at it

src/UnityExtension/Assets/Editor/GitHub.Unity/UI/PUblishView.cs
```diff
@@ -127,111 +133,111 @@ public override void OnGUI()
...
-            GUILayout.BeginHorizontal();
+            EditorGUI.BeginDisabledGroup(isLoading || isBusy);
             {
-                GUILayout.BeginVertical();
+                GUILayout.BeginHorizontal();
                 {
-                    GUILayout.Label("Owner");
+                    GUILayout.BeginVertical();
+                    {
+                        GUILayout.Label("Owner");
```
shana: Move this string to a const while you're at it

src/UnityExtension/Assets/Editor/GitHub.Unity/UI/PUblishView.cs
StanleyGoldman: I don't have a good reason for using a different flag, mental clarity if nothing else...

src/UnityExtension/Assets/Editor/GitHub.Unity/UI/PUblishView.cs
```diff
@@ -91,6 +98,8 @@ private void PopulateView()
 
                             owners = owners.Union(organizationLogins).ToArray();
                         });
+                    }).Finally(task => {
+                        isBusy = false;
```
shana: Oh, this needs to be `FinallyInUI`, serialized ui data should only be touched in thread

src/UnityExtension/Assets/Editor/GitHub.Unity/UI/PUblishView.cs
```diff
@@ -127,111 +136,111 @@ public override void OnGUI()
...
+                            }
 
-                        GitClient.RemoteAdd("origin", repository.CloneUrl)
-                                 .Then(GitClient.Push("origin", Repository.CurrentBranch.Value.Name))
-                                 .ThenInUI(Parent.Finish)
-                                 .Start();
-                    }, organization);
+                            GitClient.RemoteAdd("origin", repository.CloneUrl)
+                                     .Then(GitClient.Push("origin", Repository.CurrentBranch.Value.Name))
+                                     .ThenInUI(Parent.Finish)
```
shana: Before doing this, `isBusy` should be reset to false. We shouldn't be calling `Parent.Finish` here anyway, we should be calling `Finish` (which calls `Parent.Finish`), and then we can reset the busy flag there.

src/UnityExtension/Assets/Editor/GitHub.Unity/UI/PUblishView.cs
StanleyGoldman: Hmm, the more I look at it, I realize this `Finally` is incorrect in other ways..
The `GetCurrentUser` callback has two different execution paths, one is synchronous the other is async.

https://github.com/github-for-unity/Unity/blob/0e489246486557981c314f7c0c82a65bc65fe1b3/src/UnityExtension/Assets/Editor/GitHub.Unity/UI/PublishView.cs#L73-L103

src/UnityExtension/Assets/Editor/GitHub.Unity/UI/PUblishView.cs
StanleyGoldman: > Before doing this, isBusy should be reset to false.

😕 I dont understand what you mean by this

> We shouldn't be calling Parent.Finish here anyway, we should be calling Finish (which calls Parent.Finish), and then we can reset the busy flag there.

👍 But the rest of this I got

src/UnityExtension/Assets/Editor/GitHub.Unity/UI/PUblishView.cs
shana: aaaah yes we're not in the new task system here. bummer. Ok we can't use `Finally` here, it's on the wrong thread. Let's set `isBusy = false` in all the return points inside the callbacks for now and then talk about how this API should actually look like separately.

src/GitHub.Api/Application/ApiClient.cs
```diff
@@ -174,16 +182,15 @@ public ApiClient(UriString hostUrl, IKeychain keychain, IGitHubClient githubClie
...
-        private async Task CreateRepositoryInternal(NewRepository newRepository, Action<Octokit.Repository, Exception> callback, string organization)
+        private async Task<Octokit.Repository> CreateRepositoryInternal(NewRepository newRepository, string organization)
         {
             try
             {
                 logger.Trace("Creating repository");
 
                 if (!await EnsureKeychainLoaded())
                 {
-                    callback(null, new Exception("Keychain Not Loaded"));
```
shana: You need to throw here to have the same behaviour as before

StanleyGoldman: I mean, the user shouldn't reach here without having their Keychain loaded.
So i guess this code being here is more of a formality...

It's something I hope to clean up as we..
1. Get proper tasks around the API
2. Figure out a clean way to handle authentication errors from the API and Git

shana: Right, yeah, kinda need to rewrite this whole thing...

PR #198, opened 2017-08-16
Fix to load the PublishView faster

StanleyGoldman:
Depends on
- [x] #197

### DIFF
--- a/src/GitHub.Api/Application/ApiClient.cs
+++ b/src/GitHub.Api/Application/ApiClient.cs
@@ -283,6 +283,7 @@ namespace GitHub.Unity
 
                 logger.Trace("LoadKeychainInternal: Loading");
 
+                //TODO: ONE_USER_LOGIN This assumes only ever one user can login
                 var uriString = keychain.Connections.First().Host;
                 var keychainAdapter = await keychain.Load(uriString);
 
--- a/src/GitHub.Api/Application/ApiClient.cs
+++ b/src/GitHub.Api/Application/ApiClient.cs
@@ -78,6 +78,12 @@ namespace GitHub.Unity
             callback(user);
         }
 
+        public async Task GetCurrentUserAndOrganizations(Action<Octokit.User, IList<Organization>> callback)
+        {
+            Guard.ArgumentNotNull(callback, "callback");
+            await GetUsersAndOrganizationInternal(callback);
+        }
+
         public async Task Login(string username, string password, Action<LoginResult> need2faCode, Action<bool, string> result)
         {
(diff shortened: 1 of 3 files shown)


src/GitHub.Api/Application/ApiClient.cs
```diff
@@ -262,6 +268,23 @@ public ApiClient(UriString hostUrl, IKeychain keychain, IGitHubClient githubClie
...
+                return;
+            }
+
+            var currentUserInternal = GetCurrentUserInternal();
+            var organizationInternal = GetOrganizationInternal();
+
+            currentUserInternal.Start(TaskScheduler.Current);
+            organizationInternal.Start(TaskScheduler.Current);
+
+            callback(await currentUserInternal,await organizationInternal);
```
StanleyGoldman: This earns me a nice Unity hang... what am I doing wrong here?

src/UnityExtension/Assets/Editor/GitHub.Unity/UI/PUblishView.cs
```diff
@@ -67,33 +68,21 @@ private void PopulateView()
...
-                            if (organizations == null)
-                            {
-                                Logger.Warning("Unable to get list of organizations");
-                                return;
-                            }
-
-                            Logger.Trace("Loaded {0} organizations", organizations.Count);
-
-                            var organizationLogins = organizations
+                            var organizationLogins = (organizations ?? Enumerable.Empty<Organization>())
```
shana: Instead of returning null from `GetOrganizations`, return an empty enumerable so the caller doesn't have to guess whether it can use the return value.

src/GitHub.Api/Application/ApiClient.cs
```diff
@@ -262,27 +269,27 @@ public ApiClient(UriString hostUrl, IKeychain keychain, IGitHubClient githubClie
...
                 {
-                    logger.Trace("EnsureKeychainLoaded: Has keys does not need load");
+                    logger.Trace("LoadKeychainInternal: Has keys does not need load");
                     return true;
                 }
 
-                logger.Trace("EnsureKeychainLoaded: Loading");
+                logger.Trace("LoadKeychainInternal: Loading");
 
                 var uriString = keychain.Connections.First().Host;
```
shana: We should leave a TODO here, this is assuming there's only ever going to be one login, which isn't going to be true in the long run. We should be loading the credentials that correspond to the host of the repo we're on, or alternatively all credentials that are available for all hosts we know about.

PR #201, opened 2017-08-18
Fix console message test to be in past tense

StanleyGoldman:
It is normal for Unity3d to display the last Console Log message received in the status bar.
Some newer users are not always familiar with this and the user of active tense gives the impression that some ongoing process has yet to finish. [link](https://github.com/github-for-unity/Unity/issues/185#issuecomment-322211970)


src/UnityExtension/Assets/Editor/GitHub.Unity/EntryPoint.cs
```diff
@@ -53,7 +53,7 @@ private static void Initialize()
                     Logging.Error(ex, "Error rotating log files");
                 }
 
-                Debug.Log("Initializing GitHub for Unity log file: " + logPath);
```
shana: Oooh, this is an excellent chance to combine this message with the one above so we don't log into the console twice, there was someone complaining about that. How about

`Debug.LogFormat("Initialized GitHub for Unity version {0}. Log file: {1}", ApplicationInfo.Version, logPath);`?

PR #215, opened 2017-08-21
Fix to use correct remote name in Pull/Push changes dialog

### LINKED ISSUE #214: Pull/Push popup always shows 'placeholder' when remote is known
img

StanleyGoldman:
Fixes #214


src/UnityExtension/Assets/Editor/GitHub.Unity/UI/HistoryView.cs
```diff
@@ -351,6 +349,7 @@ public void OnEmbeddedGUI()
                 var isPublished = Repository.CurrentRemote.HasValue;
                 if (isPublished)
                 {
+                    var currentRemote = Repository.CurrentRemote.Value.Name;
```
shana: We actually want to cache both `isPublished` and `currentRemote` in serialized fields and update them in `MaybeUpdateData` so that the UI doesn't get confused if this information changes in a background thread while the UI is in the middle of a layout/repaint phase.

PR #220, opened 2017-08-22
Fix garbled characters in history view with non-ascii characters (fixes #124)(fixes #35)

### LINKED ISSUE #35: ProcessManagerIntegrationTests RussianLogEntriesTest no longer passes

### LINKED ISSUE #124: Non-ASCII character logs and file names are garbled in the history view and changes view.

shiena:
fixes #124 
fixes #35 

recreate #136

### DIFF
--- a/src/GitHub.Api/Git/Tasks/GitCommitTask.cs
+++ b/src/GitHub.Api/Git/Tasks/GitCommitTask.cs
@@ -15,7 +15,7 @@ namespace GitHub.Unity
             Guard.ArgumentNotNullOrWhiteSpace(message, "message");
 
             Name = TaskName;
-            arguments = "commit ";
+            arguments = "-c i18n.commitencoding=utf8 commit ";
             arguments += String.Format(" -m \"{0}\"", message);
             if (!String.IsNullOrEmpty(body))
                 arguments += String.Format(" -m \"{0}\"", body);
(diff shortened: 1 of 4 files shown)


shana:
celebrate

PR #235, opened 2017-08-26
Fix script for opening terminal on mac

### LINKED ISSUE #234: Shell opened from command line menu does not move to project folder
### Operating System

OS: macOS Sierra 10.12.5
Unity: 2017.1.0p4
GitHub for Unity: v0.18-alpha

https://github.com/github-for-unity/Unity/blob/b7a04efb85ddc9976c2dd6ac4bb776df0b0a68e8/src/GitHub.Api/OutputProcessors/ProcessManager.cs#L81-L91

Although the script refers to `$GHU_WORKINGDIR` and `$GHU_FULLPATH`, it seems that Terminal.app does not inherit the environment variable to the script.
Since these two variables are set by `gitEnvironment.Configure` , it is thought that they should be expanded afterwards.

shiena:
### Description of the Change

Terminal.app does not inherit environment variables.
Therefore, expand environmet variables in the script.

### Applicable Issues

fixes #234

### DIFF
--- a/src/GitHub.Api/OutputProcessors/ProcessManager.cs
+++ b/src/GitHub.Api/OutputProcessors/ProcessManager.cs
@@ -71,6 +71,7 @@ namespace GitHub.Unity
             if (environment.IsWindows)
             {
                 startInfo.FileName = "cmd";
+                gitEnvironment.Configure(startInfo, workingDirectory);
             }
             else if (environment.IsMac)
             {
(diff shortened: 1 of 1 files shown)


shana:
Weird, my terminal app inherits all environment variables except for path (which is explicitly overridden when it launches). I wonder what the difference is...¯\_(ツ)_/¯

shiena:
@shana 
Using the default login shell will inherit the environment variables.
However, using other command does not seem to inherit it.
Even if you set `/bin/bash`.
The Atlassian SourceTree is working fine, seems to be using osascript.

image

PR #249, opened 2017-08-31
Refatoring to remove unused code

StanleyGoldman:
There is a lot of unused code in our views.
As I'm copying chunks of the `SettingsView` and making changes to the `Window`, it makes sense to try to trim it back some.


StanleyGoldman:
#252 Depends on this

StanleyGoldman:
#261 Depends on this

src/UnityExtension/Assets/Editor/GitHub.Unity/UI/SettingsView.cs
```diff
@@ -792,44 +517,5 @@ private void OnLoggingSettingsGui()
...
             EditorGUI.EndDisabledGroup();
         }
-
-        private void ResetInitDirectory()
-        {
-            initDirectory = Utility.UnityProjectPath;
-            GUIUtility.keyboardControl = GUIUtility.hotControl = 0;
-        }
-
-        private void ForceUnlockFile(object obj)
```
evilshana: We should make sure we track this in an issue for fixing, we'll need to do this as part of the initial file locking support.

StanleyGoldman: Who are you?

PR #250, opened 2017-08-31
Refactoring fields of Window for clarity

StanleyGoldman:
- There is an enum called `SubTab` with members `History`, `Changes`, etc.
- There are several sub classes of `SubView`: `HistoryView`, `BranchesView`, etc.

`Window` has a member `HistoryTab` et al. which is an instance of `HistoryView` et al.
It's rather confusing, I'm renaming them to be the same as the type, which isn't great but better than the `-Tab` suffix I felt.
The alternative would be to name them `History` et al.


StanleyGoldman:
#252 Depends on this

StanleyGoldman:
#261 Depends on this

StanleyGoldman:
Rebased

PR #252, opened 2017-08-31
A dedicated view to initialize the project

### LINKED ISSUE #251: Create a dedicated view to initialize the project
We need a dedicated view to handle project initialization and possibly the "first time git initialization".
It should provide the following features.

- A form to handle the git username and email.
  - Users may not have a this data on their machines already
  - This form should allow for validation errors
- A form that presents their git installation path
  - Users may want to change this path
  - This form should allow for validation errors
- A final button to initialize the project

The final work is being done in: 
- #261 
- [branch: initialize-view-with-configuration](https://github.com/github-for-unity/Unity/tree/fixes/initialize-view-with-configuration)

StanleyGoldman:
As part of #251

Depends on:
- [x] #249 
- [x] #250
- [x] #253
- [x] #260

Before:
img

After:
img


src/UnityExtension/Assets/Editor/GitHub.Unity/UI/Window.cs.orig
```diff
@@ -0,0 +1,467 @@
+#pragma warning disable 649
```
StanleyGoldman: Ugh me.

src/UnityExtension/Assets/Editor/GitHub.Unity/UI/Window.cs
```diff
@@ -158,6 +162,12 @@ public override void OnUI()
 
             DoToolbarGUI();
 
+            if (nextTab.HasValue)
+            {
+                SetActiveTab(nextTab.Value);
```
shana: This can't be done here, SetActiveTab calls OnDisable and OnEnable on the views, and they're not set up to expect these events multiple times (OnUI is called multiple times, one for each phase of the render pass). Also these events should be called outside the UI phase in general, preferably before OnDataUpdate is called on the views (which happens before OnUI is called).

TL;DR, the order of events for child views should be "OnDisable" (the old one) -> "OnEnable" (the new one) -> "OnDataUpdate" -> "OnGUI"

src/UnityExtension/Assets/Editor/GitHub.Unity/UI/Window.cs
```diff
@@ -24,8 +25,10 @@ class Window : BaseWindow
         private const string Window_RepoBranchTooltip = "Active branch";
 
         [NonSerialized] private double notificationClearTime = -1;
+        [SerializeField] private SubTab? nextTab;
```
shana: SubTab is an enum so there's no reason to use a nullable here, just add a None option to it to know what there is no current selection.

src/UnityExtension/Assets/Editor/GitHub.Unity/UI/Window.cs
```diff
@@ -417,20 +444,34 @@ private Subview ToView(SubTab tab)
...
+
                 case SubTab.History:
                     return historyView;
+
                 case SubTab.Changes:
                     return changesView;
+
                 case SubTab.Branches:
                     return branchesView;
+
```
shana: Why the extra space? Seems like a useless change.

src/UnityExtension/Assets/Editor/GitHub.Unity/UI/Window.cs
```diff
@@ -417,20 +444,34 @@ private Subview ToView(SubTab tab)
...
             {
+                case SubTab.InitProject:
+                    return initProjectView;
+
                 case SubTab.History:
                     return historyView;
+
                 case SubTab.Changes:
                     return changesView;
+
```
shana: No need for this change

src/UnityExtension/Assets/Editor/GitHub.Unity/UI/Window.cs
```diff
@@ -417,20 +444,34 @@ private Subview ToView(SubTab tab)
         {
             switch (tab)
             {
+                case SubTab.InitProject:
+                    return initProjectView;
+
                 case SubTab.History:
                     return historyView;
+
```
shana: No need for this change

PR #253, opened 2017-08-31
Removing broadmode code

StanleyGoldman:

PR #254, opened 2017-08-31
Broadmode

StanleyGoldman:
Depends on:
- [ ] #253


StanleyGoldman:
Pretty sure we are not bringing back "broadmode" functionality

PR #260, opened 2017-08-31
Fixing Window.MaybeUpdateData when Repository is null

### LINKED ISSUE #259: When Repository is null Window.MaybeUpdateData() always returns true
This is a waste before things are initialized and leads to a bad code pattern in `Window.MaybeUpdateData(){...}`

StanleyGoldman:
Fixes #259


StanleyGoldman:
#252 Depends on this

StanleyGoldman:
#261 Depends on this

PR #261, opened 2017-08-31
Adding functionality to configure git exec, name and email before project init

StanleyGoldman:
Fixes #251 

Depends on:
- [x] #252 

Which Depends on:
- [x] #249 
- [x] #250
- [x] #253
- [x] #260

Before:
img

After #252:
img

After this:
img

PR #270, opened 2017-09-02
Isolating the git user and email into a UserSettingsView

StanleyGoldman:
Part of #251
#261 depends on this

Depends on:
- [x] #273 
- [x] #277 

- There are no visible changes to the UI as a result of this change
- These changes also prevent the save of a blank name or email


src/UnityExtension/Assets/Editor/GitHub.Unity/UI/PopupWindow.cs
```diff
@@ -93,6 +93,11 @@ public override void OnSelectionChange()
             ActiveView.OnSelectionChange();
         }
 
+        public override bool IsBusy
+        {
+            get { return ActiveView.IsBusy; }
```
shana: This change is probably a glitchy merge, the property should be at the bottom

src/UnityExtension/Assets/Editor/GitHub.Unity/UI/SettingsView.cs
```diff
@@ -68,18 +69,25 @@ public override void OnEnable()
...
         {
             base.OnDisable();
+            userSettingsView.OnDisable();
             DetachHandlers(Repository);
         }
 
         public override void OnDataUpdate()
         {
             base.OnDataUpdate();
+            if (userSettingsView != null)
```
shana: When is this ever null?

src/UnityExtension/Assets/Editor/GitHub.Unity/UI/SettingsView.cs
```diff
@@ -89,9 +97,15 @@ public override void OnRepositoryChanged(IRepository oldRepository)
             Refresh();
         }
 
+        public override bool IsBusy
+        {
+            get { return isBusy || userSettingsView.IsBusy; }
```
shana: Bottom of file

src/UnityExtension/Assets/Editor/GitHub.Unity/UI/SettingsView.cs
StanleyGoldman: You are right, removing the null check

PR #272, opened 2017-09-05
Isolating the git path functionality into GitPathView

### LINKED ISSUE #61: Add functionality to verify custom git installations
If the chosen git installation is not the "pre-packaged" git, we should should be able to verify the git installation and present a warning and possibly disable certain functions based on the result.

We should check for things like:
1. Git Version
2. Git LFS Present
3. Git LFS Version

StanleyGoldman:
Fixes #61 

Part of #251

Depends on:
- [x] #273 
- [x] #281 
- [x] #287
- [x] #292

Before:
img

After #288:
img

After this:

img

image

- Save path is only enabled with path exists
- Find button no longer automatically saves and requires an explicit save to perform version check


StanleyGoldman:
#261 Depends on this

StanleyGoldman:
@donokuda I updated this pull request with images for you to see what i did...

src/UnityExtension/Assets/Editor/GitHub.Unity/UI/GitPathView.cs
```diff
@@ -0,0 +1,286 @@
...
+            gitExecHasChanged = true;
+        }
+
+        public override void OnDataUpdate()
+        {
+            base.OnDataUpdate();
+            MaybeUpdateData();
+        }
+
+        public override bool IsBusy
```
shana: This thing goes at the end of the file

src/UnityExtension/Assets/Editor/GitHub.Unity/UI/GitPathView.cs
```diff
@@ -0,0 +1,286 @@
...
+        {
+            get { return isBusy; }
+        }
+
+        public override void OnGUI()
+        {
+            // Install path
+            GUILayout.Label(GitInstallTitle, EditorStyles.boldLabel);
+
+            EditorGUI.BeginDisabledGroup(IsBusy || Parent.IsBusy);
```
shana: If we always have to check the value of the parent, we should either be setting it directly or the `IsBusy` property should be returning the aggregate status so that `IsBusy` always reflects the state of the current view and the hierarchy up the chain. Otherwise it's an opportunity to forget to do it in some place.

src/UnityExtension/Assets/Editor/GitHub.Unity/UI/GitPathView.cs
```diff
@@ -0,0 +1,286 @@
...
+                    {
+                        newGitExec = EditorGUILayout.TextField(PathToGit, newGitExec);
+                    }
+
+                    if (EditorGUI.EndChangeCheck())
+                    {
+                        CheckEnteredGitPath();
+                    }
+
+                    if (GUILayout.Button(BrowseButton, EditorStyles.miniButton, GUILayout.Width(25)))
```
shana: Width should be defined in a style (either here or in the Styles class) and not hardcoded here

src/UnityExtension/Assets/Editor/GitHub.Unity/UI/GitPathView.cs
```diff
@@ -0,0 +1,286 @@
...
+                    if (GUILayout.Button(GitInstallFindButton, GUILayout.ExpandWidth(false)))
+                    {
+                        GUI.FocusControl(null);
+                        isBusy = true;
+
+                        newGitExec = gitExec;
+                        CheckEnteredGitPath();
+
+                        new ProcessTask<NPath>(Manager.CancellationToken, new FirstLineIsPathOutputProcessor())
+                            .Configure(Manager.ProcessManager, Environment.IsWindows ? "where" : "which", "git")
```
shana: `FindExecTask` that does this type of executable location via where/which

src/UnityExtension/Assets/Editor/GitHub.Unity/UI/GitPathView.cs
```diff
@@ -0,0 +1,286 @@
...
+                                    else
+                                    {
+                                        Logger.Error("FindGit Failed Path:{0}", path);
+                                    }
+                                }
+
+                                if (success)
+                                {
+                                    newGitExec = path;
+                                    CheckEnteredGitPath();
```
shana: This block of code should probably be merged into the `if (success)` above?

src/UnityExtension/Assets/Editor/GitHub.Unity/UI/GitPathView.cs
```diff
@@ -0,0 +1,286 @@
...
+                }
+            }
+            EditorGUI.EndDisabledGroup();
+        }
+
+        private void MaybeUpdateData()
+        {
+            if (gitExecHasChanged)
+            {
+                if (Environment != null)
```
shana: This is never not null since it comes directly from the cache and is automatically created on first access

src/UnityExtension/Assets/Editor/GitHub.Unity/UI/SettingsView.cs
```diff
@@ -70,18 +78,25 @@ public override void OnEnable()
...
             base.OnDisable();
+            gitPathView.OnDisable();
             DetachHandlers(Repository);
         }
 
         public override void OnDataUpdate()
         {
             base.OnDataUpdate();
+
+            if (gitPathView != null)
```
shana: Is this ever null here? We're not doing this check in any other method so it feels like it never is.

src/UnityExtension/Assets/Editor/GitHub.Unity/UI/GitPathView.cs
StanleyGoldman: I had the same feelings, but I couldn't figure out a way to do it that I was truly happy with.
Right now `IsBusy` represents self and children.

src/UnityExtension/Assets/Editor/GitHub.Unity/UI/GitPathView.cs
StanleyGoldman: 👍

src/UnityExtension/Assets/Editor/GitHub.Unity/UI/GitPathView.cs
StanleyGoldman: 👍

src/UnityExtension/Assets/Editor/GitHub.Unity/UI/GitPathView.cs
StanleyGoldman: 👍

src/UnityExtension/Assets/Editor/GitHub.Unity/UI/GitPathView.cs
StanleyGoldman: 👍

src/UnityExtension/Assets/Editor/GitHub.Unity/UI/GitPathView.cs
StanleyGoldman: 👍

src/UnityExtension/Assets/Editor/GitHub.Unity/UI/SettingsView.cs
StanleyGoldman: 👍

PR #273, opened 2017-09-05
Adding IsBusy to IView

StanleyGoldman:
In order for future `IView`'s to communicate their `IsBusy` status to their parents.


StanleyGoldman:
#272 #270 depends on this

PR #275, opened 2017-09-05
Refactor views to update settings in MaybeUpdateData

StanleyGoldman:


src/UnityExtension/Assets/Editor/GitHub.Unity/UI/SettingsView.cs
```diff
@@ -139,8 +144,20 @@ public override void OnGUI()
             GUILayout.EndScrollView();
         }
 
+        public override void InitializeView(IView parent)
+        {
+            base.InitializeView(parent);
+            usageTracker = Manager != null && Manager.UsageTracker != null ? Manager.UsageTracker : null;
```
shana: This is not needed, the Manager is always guaranteed to exist

shana: UsageTracker is never null either so we don't have to worry about that

PR #281, opened 2017-09-05
Functionality to send commands for and parse git and git lfs version

StanleyGoldman:
Depends on:
- [x] #280


StanleyGoldman:
Hmm, I can't say I am expecting these tests to fail... they work locally.

src/GitHub.Api/Git/GitClient.cs
```diff
@@ -148,9 +152,39 @@ private Task<NPath> LookForPortableGit()
...
         }
 
-        public bool ValidateGitInstall(NPath path)
+        public ITask<ValidateGitInstallResult> ValidateGitInstall(string path)
         {
-            return path.FileExists();
+            if (!path.ToNPath().FileExists())
+            {
+                return new FuncTask<ValidateGitInstallResult>(cancellationToken, 
+                    () => new ValidateGitInstallResult(false, null, null));
```
shana: This should probably be one of those Completed tasks that we never know the name of?

src/GitHub.Api/Git/GitClient.cs
```diff
@@ -148,9 +152,39 @@ private Task<NPath> LookForPortableGit()
...
+            }
+
+            Version gitVersion = null;
+            Version gitLfsVersion = null;
+
+            var gitVersionTask = new GitVersionTask(cancellationToken);
+            gitVersionTask.Configure(processManager, path, 
+                gitVersionTask.ProcessArguments, environment.RepositoryPath);
```
shana: You can do `var gitVersionTask = new GitVersionTask(cancellationToken).Configure(...)`, Configure returns the task.

src/GitHub.Api/Git/GitClient.cs
```diff
@@ -148,9 +152,39 @@ private Task<NPath> LookForPortableGit()
...
+
+            var gitLfsVersionTask = new GitLfsVersionTask(cancellationToken);
+            gitLfsVersionTask.Configure(processManager, path,
+                gitLfsVersionTask.ProcessArguments, environment.RepositoryPath);
+
+            return gitVersionTask
+                .Then((result, version) => { gitVersion = version; })
```
shana: Nitpicking the braces and semicolon here, no real need for them. :nitpick: :nitpick:

src/GitHub.Api/Git/Tasks/GitLfsVersionTask.cs
```diff
@@ -0,0 +1,19 @@
...
+        private const string TaskName = "git lfs version";
+
+        public GitLfsVersionTask(CancellationToken token, IOutputProcessor<Version> processor = null)
+            : base(token, processor ?? new LfsVersionOutputProcessor())
+        {
+            Name = TaskName;
+        }
+
+        public override string ProcessArguments { get { return "lfs version"; } }
+        public override TaskAffinity Affinity { get { return TaskAffinity.Concurrent; } }
```
shana: This is the default concurrency, just fyi (being explicit about it is fine)

src/GitHub.Api/OutputProcessors/LfsVersionOutputProcessor.cs
```diff
@@ -0,0 +1,22 @@
...
+            if (String.IsNullOrEmpty(line))
+                return;
+
+            var gitVersion = "git-lfs/";
+            if (line.StartsWith(gitVersion))
+            {
+                line = line.Substring(gitVersion.Length, line.IndexOf(" ", StringComparison.InvariantCultureIgnoreCase) - gitVersion.Length);
+                var strings = line.Split(new[] { "." }, StringSplitOptions.None);
+
+                RaiseOnEntry(new Version(Int32.Parse(strings[0]), Int32.Parse(strings[1]), Int32.Parse(strings[2])));
```
shana: This can definitely throw if the format changes. Maybe a regex would be better here for flexibility? At least some TryParse calls for not throwing

src/GitHub.Api/OutputProcessors/VersionOutputProcessor.cs
```diff
@@ -0,0 +1,23 @@
...
+            if (String.IsNullOrEmpty(line))
+                return;
+
+            var gitVersion = "git version ";
+            if (line.StartsWith(gitVersion))
+            {
+                line = line.Substring(gitVersion.Length);
+                var strings = line.Split(new[] { "." }, StringSplitOptions.None);
+
+                RaiseOnEntry(new Version(Int32.Parse(strings[0]), Int32.Parse(strings[1]), Int32.Parse(strings[2])));
```
shana: I am pretty sure someone has already written a regex somewhere to parse the git version, but in any case something like `/git version ([\d]+)\.([\d]+)\.([\d]+)` should probably work and be resilient enough?

src/tests/UnitTests/IO/VersionOutputProcessorTests.cs
```diff
@@ -0,0 +1,38 @@
...
+        {
+            TestCaseData testCase;
+
+            testCase = new TestCaseData(
+                $"git version 2.11.1.windows.1{Environment.NewLine}",
+                new Version(2, 11, 1));
+
+            testCase.SetName("Windows 2.11.1");
+            yield return testCase;
+        }
```
shana: This is where we should have lots and lots of weird-ass version strings to throw at the processor

shana: Does every git report 3 parts to the version followed by an OS string?

src/GitHub.Api/Git/GitClient.cs
StanleyGoldman: `TaskHelpers.GetCompletedTask` returns a `Task<T>` but I need `ITask<T>`. We dont have a `GetCompletedTask` method for `ITask<T>`

src/GitHub.Api/Git/GitClient.cs
StanleyGoldman: I needed to get my hands on `gitVersionTask.ProcessArguments` on the second line.

src/GitHub.Api/Git/GitClient.cs
StanleyGoldman: 👍

src/GitHub.Api/Git/Tasks/GitLfsVersionTask.cs
StanleyGoldman: Cool

src/tests/UnitTests/IO/VersionOutputProcessorTests.cs
StanleyGoldman: I added the git version and git lfs version from my mac

src/GitHub.Api/OutputProcessors/VersionOutputProcessor.cs
StanleyGoldman: Changed to a regex

src/GitHub.Api/Git/GitClient.cs
shana: Oooh! Just fix the extension method so it passes on the process arguments, in that case, that's what it's for.

src/GitHub.Api/Git/GitClient.cs
shana: So the way to do that here is to do `return new FuncTask<ValidateGitInstallResult>(TaskEx.FromResult(new ValidateGitInstallResult(false, null, null)));`

This will create a completed task for FuncTask, and when that gets added to the task continuation, it'll automatically be done and won't start up a thread at all.

shana:
Closing this one out to have them in context with the PR that actually uses these changes

PR #288, opened 2017-09-06
Adding a save button for the git path

### LINKED ISSUE #286: Manual Git path selection needs a "Save" button
Typing in the "Path to Git" textbox causes a validation on every keystroke.
That doesn't allow us a good "choke point" to validate the chosen git path.

StanleyGoldman:
Fixes #286 

Also refactoring to better control when we update the git path from the environment.

Before:
img

After:
img

- It enables when the path is changed.
- The textbox and the browse button no longer automatically attempt to save. Clicking save is required.

Depends on:
- [x] #287


shana:
Could we get a screenshot of what it looks like now? Also needs some conflict resolving.

StanleyGoldman:
Added pics to pull request

StanleyGoldman:
I made a mistake in documenting my jenga pile, this functionality was included in #272

PR #298, opened 2017-09-11
Keeping the authentication if the username doesn't match in the keychain

### LINKED ISSUE #177: Authentication validation on startup is often incorrect and drops credentials
It might just be an impression I'm getting, but I seem to lose my credentials way too often.

StanleyGoldman:
Fixes #177

Depends on:
- [x] #296


src/GitHub.Api/Authentication/Keychain.cs
```diff
@@ -55,7 +55,7 @@ public IKeychainAdapter Connect(UriString host)
             KeychainAdapter keychainAdapter = FindOrCreateAdapter(host);
             var cachedConnection = connectionCache[host];
 
-            logger.Trace($@"Loading KeychainAdapter Host:""{host}"" Cached Username:""{cachedConnection.Username}""");
+            logger.Trace("Loading KeychainAdapter Host:\"{0}\" Cached Username:\"{1}\"", host, cachedConnection.Username);
```
shana: Formatting change that doesn't need to be here

PR #299, opened 2017-09-11
Fixes to keep the publish view from loading twice

StanleyGoldman:
The current way `PopulateView` is being called causes two concurrent calls to `LoadKeychain`.

PR #300, opened 2017-09-11
Switch to AuthView on Publish API Failure

StanleyGoldman:
Currently our only exposure to the API is `PublishView`. In the event of an authentication failure we need to switch the `PublishView` to the `AuthenticationView`

- If the user's authentication is not who we expect
img

-If there is no authentication present
img

- Sign In vs Publish
img

Depends on:
- [x] #295 
- [x] #298 
- [x] #299
- [x] #361


StanleyGoldman:
Do we need message to explain to the user that we need your authentication?
@shana @donokuda

donokuda:
> Do we need message to explain to the user that we need your authentication?

Yup, I'm 👍 in favor of explaining why they're taken to the authentication screen. Any strong feelings about checking first if they're signed in before displaying the publish form? If they're not logged in, then we could show them the sign-in form with a little error up on top:



I can push up an example branch if that helps!

StanleyGoldman:
Cool. Thanks for the draft.
I think I got it, I'll let you know if I need more.

donokuda:
Added back the header styles:

src/UnityExtension/Assets/Editor/GitHub.Unity/UI/PopupWindow.cs
```diff
@@ -14,38 +14,137 @@ public enum PopupViewType
...
+                }, exception => {
+                    Logger.Trace("User required validation opening AuthenticationView");
+
+                    string message = null;
+                    string username = null;
+
+                    var usernameMismatchException = exception as TokenUsernameMismatchException;
+                    if (usernameMismatchException != null)
+                    {
+                        message = "Your credentials need to be refreshed";
```
shana: Been thinking about this some more and talking to some people about what messages should say, and we should probably explain a bit more why we're doing this.

How about:

We've detected that your stored credentials are out of sync with your current user. This can happen if you have signed in to git outside of Unity. Sign in again to refresh your credentials.

src/UnityExtension/Assets/Editor/GitHub.Unity/UI/PopupWindow.cs
```diff
@@ -14,38 +14,137 @@ public enum PopupViewType
...
+
+                    var keychainEmptyException = exception as KeychainEmptyException;
+                    if (keychainEmptyException != null)
+                    {
+                        message = "We need you to authenticate first";
+                    }
+
+                    if (usernameMismatchException == null && keychainEmptyException == null)
+                    {
+                        message = "There was an error validating your account";
```
shana: All of these messages should be in const fields somewhere so we can localize them later

shana: I feel like this message is a bit unhelpful, especially given that it's the first thing they see right after clicking the button. When does this happen and is this our fault or the users'? If it's our fault, then we should followup with "Please sign in again" to tell them how to proceed here.

src/UnityExtension/Assets/Editor/GitHub.Unity/UI/PopupWindow.cs
StanleyGoldman: That sounds much better.

src/UnityExtension/Assets/Editor/GitHub.Unity/UI/PopupWindow.cs
StanleyGoldman: It actually hasn't happened to me yet, so it would probably be best to leave the original exception message there instead.

src/UnityExtension/Assets/Editor/GitHub.Unity/UI/PublishView.cs
```diff
@@ -35,7 +39,7 @@ class PublishView : Subview
         [NonSerialized] private IApiClient client;
```
shana: Not directly related to this line, but some lines above the `IList<Organization>` field needs fixing. It isn't serializable by Unity, it needs to be an instance of a thing. Also, `Organization` is probably not very serializable either, given what it contains, so we probably need a type of our own for this and serialize it as `List<Thing>`

src/GitHub.Api/Application/ApiClient.cs
```diff
@@ -219,66 +220,69 @@ public ApiClient(UriString hostUrl, IKeychain keychain, IGitHubClient githubClie
             }
         }
 
-        private async Task<IList<Organization>> GetOrganizationInternal()
+        private async Task GetOrganizationInternal(Action<IList<Organization>> onSuccess, Action<Exception> onError = null)
```
shana: Related to https://github.com/github-for-unity/Unity/pull/300#discussion_r146317238, `IList<Organization>` is not a very useful type, especially since it won't be serializable by Unity (meaning we need to get a `ToArray()` here to get it and then we need to do a `ToList()` in the UI to serialize it, so copying the data twice for no big gains)

Also, `Organization` is not easily serializable and has too much junk in it that we don't need, we should wrap this in an object we control and not expose Octokit models directly to the outside.

StanleyGoldman: I opened up #386 to resolve this issue

StanleyGoldman:
@donokuda I feel like the current auth view does not match the publish view. I think it was different before because you would completely leave the auth view and then separately enter the publish view. After these changes that will no longer be the case, one will flow into another.

Current Auth View:
img

Current Publish View:
img

Maybe it's just the lack of an invertocat that makes me feel like it's missing.

donokuda:
Here's where this PR is at right now. It's ready for some feedback:

Sign in view with warning:

screen shot 2017-10-27 at 9 18 45 am

2FA view (is there a way to suppress that error if this is the first time viewing this screen?)

screen shot 2017-10-27 at 9 24 19 am

I also noticed when you hit "back" from the 2FA, it will still persist the error message even though it's not relevant to the screen before. Is there a way to prevent that from happening?

screen shot 2017-10-27 at 9 30 21 am

shana:
I like this UI *a lot*, those message box styles rock!

> I also noticed when you hit "back" from the 2FA, it will still persist the error message even though it's not relevant to the screen before. Is there a way to prevent that from happening?

We probably need to call Repaint/Redraw on the window to refesh the UI here

I wonder how we handle the Esc key. We should probably close the window and cancel the whole thing.

src/UnityExtension/Assets/Editor/GitHub.Unity/UI/PublishView.cs
```diff
@@ -89,59 +93,78 @@ public override void InitializeView(IView parent)
...
-                            var organizationLogins = organizations
-                                .OrderBy(organization => organization.Login).Select(organization => organization.Login);
+                    var shouldProceed = EditorUtility.DisplayDialog(AuthenticationChangedTitle,
+                        string.Format(AuthenticationChangedMessageFormat,
+                            tokenUsernameMismatchException.CachedUsername,
+                            tokenUsernameMismatchException.CurrentUsername), AuthenticationChangedProceed, AuthenticationChangedLogout);
+
+                    if (shouldProceed)
+                    {
+                        //Proceed as current user
```
StanleyGoldman: Still need to work out what happens here..

src/UnityExtension/Assets/Editor/GitHub.Unity/UI/PublishView.cs
```diff
@@ -89,59 +93,78 @@ public override void InitializeView(IView parent)
...
 
-                            owners = new[] { OwnersDefaultText, username }.Union(organizationLogins).ToArray();
+                    }
+                    else
+                    {
+                        //Logout current user and try again
```
StanleyGoldman: And here..

StanleyGoldman:
Actually... the more I look at it this code will never run, `PopupWindow` and `AuthenticationView` already has us covered on this one.

https://github.com/github-for-unity/Unity/blob/b108e06d1d4c482aef37a1e96ece7c504dfd05ca/src/UnityExtension/Assets/Editor/GitHub.Unity/UI/PublishView.cs#L126-L147

StanleyGoldman:
Fixes #414

Issue #203, opened 2017-08-18
Consolidate initialization message to a single line


shana:
Currently we're showing two messages in the Unity console when it starts up, one with the plugin version and another with the log file. We should just do it in one message so we don't spam the log as much.

shana:
See https://github.com/github-for-unity/Unity/pull/201#discussion_r133999327

AkashGutha:
@shana i'll grab this. :)

PR #308, opened 2017-09-13
Consolidating initializing message to a single line

AkashGutha:
### Requirements

Consolidating the logging info into a single line.

### Description of the Change

Still a work in progress. So, the idea is to implement a combined logger.
is that the aim of #203

### Alternate Designs

If not we can just remove one of the log lines

### Benefits

Having a combined logger will make things simpler going forward.

### Possible Drawbacks

I possibly don't see any.

### Applicable Issues

#203

### DIFF
--- a/src/UnityExtension/Assets/Editor/GitHub.Unity/EntryPoint.cs
+++ b/src/UnityExtension/Assets/Editor/GitHub.Unity/EntryPoint.cs
@@ -34,11 +34,10 @@ namespace GitHub.Unity
 
             // this will initialize ApplicationManager and Environment if they haven't yet
             var logPath = Environment.LogPath;
+            var logInfo = "";
 
             if (ApplicationCache.Instance.FirstRun)
             {
-                Debug.Log("Initialized GitHub for Unity version " + ApplicationInfo.Version);
-
                 var oldLogPath = logPath.Parent.Combine(logPath.FileNameWithoutExtension + "-old" + logPath.ExtensionWithDot);
                 try
                 {
@@ -53,10 +52,15 @@ namespace GitHub.Unity
                     Logging.Error(ex, "Error rotating log files");
                 }
 
-                Debug.Log("Initialized GitHub for Unity log file: " + logPath);
+                logInfo = "Initialized GitHub for Unity verison " + ApplicationInfo.Version + " and Unity log file: " + logPath;
+                Debug.Log(logInfo);
             }
+
             Logging.LogAdapter = new FileLogAdapter(logPath);
-            Logging.Info("Initializing GitHub for Unity version " + ApplicationInfo.Version);
+            if (logInfo != "")
+            {
+                Logging.Info(logInfo);
+            }
 
             ApplicationManager.Run(ApplicationCache.Instance.FirstRun);
         }
--- a/src/GitHub.Api/NewTaskSystem/ProcessTask.cs
+++ b/src/GitHub.Api/NewTaskSystem/ProcessTask.cs
@@ -288,8 +288,8 @@ namespace GitHub.Unity
                     if (outputProcessor != null)
                         result = outputProcessor.Result;
 
-                    if (result == null && typeof(T) == typeof(string))
-                        result = (T)(object)(Process.StartInfo.CreateNoWindow ? "Process finished" : "Process running");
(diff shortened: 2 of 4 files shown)


StanleyGoldman:
Hey @AkashGutha thanks for participating! This task is a really simple one.
I'll leave some comments to give you some direction...

StanleyGoldman:
Okay, I understand why you might be confused, you are looking at two different types of loggers there...

`Debug.Log` is the standard Unity console log.
`Logging.Info` is our custom logger that writes to the log file on disk.

For #203 we are looking to combine the messages we send to the Unity console log.

src/UnityExtension/Assets/Editor/GitHub.Unity/EntryPoint.cs
```diff
@@ -37,6 +37,7 @@ private static void Initialize()
 
             if (ApplicationCache.Instance.FirstRun)
             {
+                // TODO: should be replaced with a common logger ?
                 Debug.Log("Initialized GitHub for Unity version " + ApplicationInfo.Version);
```
StanleyGoldman: Combine the message we send to this `Debug.Log`..

src/UnityExtension/Assets/Editor/GitHub.Unity/EntryPoint.cs
```diff
@@ -52,10 +53,11 @@ private static void Initialize()
                 {
                     Logging.Error(ex, "Error rotating log files");
                 }
-
+                // TODO: should be replaced with a common logger ?
                 Debug.Log("Initialized GitHub for Unity log file: " + logPath);
```
StanleyGoldman: ... with the message we send to this `Debug.Log`

src/UnityExtension/Assets/Editor/GitHub.Unity/EntryPoint.cs
```diff
@@ -52,10 +53,11 @@ private static void Initialize()
...
             }
             Logging.LogAdapter = new FileLogAdapter(logPath);
+            // TODO: should be replaced with a common logger ?
             Logging.Info("Initializing GitHub for Unity version " + ApplicationInfo.Version);
```
StanleyGoldman: This one should be left alone.

AkashGutha:
Also, I wanted to raise a discussion.

Why not use interpolated strings for the entire project?

StanleyGoldman:
### TLDR;
`GitHub.Unity` is restricted to C# 4 language features. You cannot use string interpolation, `async`/`await` and other C# 6.0 goodies in `GitHub.Unity`.

### Long Form Answer
This is a part of how our project and solution is configured and that is a very good question...
I want you to notice two things...

- Language Configuration
  - `GitHub.*` is set to c# 6 language features
  - `GitHub.Unity` is restricted to C# 4 language features
- Folder Structure
  - `GitHub.Api` and `GitHub.Logging` can be found in `/src/`
  - `GitHub.Unity` can be found in `/src/UnityExtension/Assets/Editor`

This allows us to open `/src/UnityExtension` as a project in Unity.
We mostly do this to quickly prototype UI changes at run time in Unity.

StanleyGoldman:
FYI we are talking about these two messages

img

AkashGutha:
I'm a bit curious as to what would be the default log message if it wasn't the firsts run?

StanleyGoldman:
That `FirstRun` section only runs when you first open Unity.
If you had that code outside that `if` statement, it would run every time you pushed play and stop in Unity.

AkashGutha:
I guess that should be fixed. 
Having troubles setting up build. 
I'll post once done.

src/UnityExtension/Assets/Editor/GitHub.Unity/EntryPoint.cs
```diff
@@ -53,10 +52,15 @@ private static void Initialize()
...
+                logInfo = "Initialized GitHub for Unity verison " + ApplicationInfo.Version + " and Unity log file: " + logPath;
+                Debug.Log(logInfo);
             }
+
             Logging.LogAdapter = new FileLogAdapter(logPath);
-            Logging.Info("Initializing GitHub for Unity version " + ApplicationInfo.Version);
+            if (logInfo != "")
+            {
+                Logging.Info(logInfo);
+            }
```
StanleyGoldman: Please revert this chunk. We were only looking to change the log message above.

PR #309, opened 2017-09-13
Refactor to use IOutputProcessor instead of BaseOutputListProcessor

StanleyGoldman:


shana:
Why is this refactor being done? `BaseOutputListProcessor` is a helper class made just so we don't have to repeatedly type out the two types, is it causing issues?

StanleyGoldman:
The intent was to change the signatures of `IGitClient` and ripple that through the Tasks for those commands.

1. It makes `IGitClient` a bit more uniform in it's arguments.
2. It reduces the amount of items I need to make public in #310

shana:
So using the interfaces directly can potentially introduce coding errors that are hard to track. `IOutputProcessor<KeyValuePair<string, object>, List<KeyValuePair<string, object>>>` is not only hard to type but also hard to read, and if the types on both sides are not the same, the compiler won't warn you and it will cause cast exceptions all over the place, which is why the `BaseOutputListProcessor<T> : IOutputProcessor<T, List<T>>` class exists - to make sure you only pass one type and it unfolds the declaration correctly. I'm very weary of changing this to use the raw interface signature and potentially open ourselves to typos and hard to track error conditions.

StanleyGoldman:
Okay, those are fair points. I'm going to close this task and do some rework in #310

PR #313, opened 2017-09-13
Disabling Authenticate menu item when appropriate

### LINKED ISSUE #306: Authenticate menu option should disappear after successfully authenticating with GitHub
### Description
The GitHub menu always shows the Authenticate option.

### Steps to Reproduce

1. Go to GitHub > Authenticate
2. Sign in with GitHub
3. Go to GitHub > Authenticate again

**Expected behavior:** 
The Authenticate option should not be shown in the GitHub menu.

authenticate option

StanleyGoldman:
Fixes #306 

After you add a menu item, it cannot be hidden.
All you can do is disable it.


shana:
This menu is actually kinda just a debug menu and not something we want to ship for realsies, and it should always be enabled to make sure we (and users) always have access to the UI in case we break something else.

PR #319, opened 2017-09-14
Preventing initialize when name & email is not present

### LINKED ISSUE #315: Global .gitconfig with name/email is required
### Prerequisites

- Be sure to run with tracing enabled to capture runtime details in the log file
- Include the log file in the PR.
  - On Windows, the extension log file is at `%LOCALAPPDATA%\GitHubUnity\github-unity.log`
[github-unity-log.txt](https://github.com/github-for-unity/Unity/files/1302222/github-unity-log.txt)

### Description

Hitting the commit button deletes the commit message, but no commit shows up in the history.

### Steps to Reproduce

1. Did a fresh clone of a unity/github testing repository, it was uploaded from PC1 and cloned on PC2 using Sourcetree.
2. Open the repo/project in Unity, import github for unity v0.20
3. gh4u thinks some files have been deleted and created, e.g. the .gitignore file. Some other git tools show it like that when a file has been renamed!?
4. Select the deleted and the created gitignore file and try to commit, but no commit shows up in the history

StanleyGoldman:
img

Fixes #315 

Depends on:
- [x] #270 
- [x] #252


shana:
So functionally this is fine, but it's very user unfriendly. If we're requiring a specific user action, we shouldn't present it in the same way as we do errors, users will think it's an error. We should also provide an easy way to go to the place where the user is expected to fill out the needed information - i.e., if we say "Fill out x in this window", we should have a link or button that sends them to it, or we should specify how to get to it.

This is btw why I would prefer not to do these UI changes until next week, because it's just another piece of temporary UI that we will have to fix :/

StanleyGoldman:
What I don't like about this is the odd repetition of this functionality...

https://github.com/github-for-unity/Unity/blob/5e02c3984049837beea82313d16c8f6bdddf601e/src/UnityExtension/Assets/Editor/GitHub.Unity/UI/InitProjectView.cs#L139-L166

https://github.com/github-for-unity/Unity/blob/5e02c3984049837beea82313d16c8f6bdddf601e/src/UnityExtension/Assets/Editor/GitHub.Unity/UI/UserSettingsView.cs#L118-L132

I contemplated making `InitProjectView` use `UserSettingsView` but I didn't come up with anything I liked... What do you think @shana?

StanleyGoldman:
Maybe a combined method in `GitClient` since it's such a common thing to query for both values?

StanleyGoldman:
Yep, that took care of things nicely.
I'm glad we had this chat...

shana:
🦆

src/UnityExtension/Assets/Editor/GitHub.Unity/UI/UserSettingsView.cs
```diff
@@ -100,23 +111,26 @@ private void MaybeUpdateData()
...
                     {
-                        if (success && !String.IsNullOrEmpty(user.Name))
-                        {
-                            cachedUser = user;
-                            userDataHasChanged = true;
-                            Redraw();
-                        }
-                    })
-                    .Start();
+                        GitClient.GetConfigUserAndEmail().FinallyInUI((success, ex, strings) => {
```
shana: This is the same code that RepositoryManager is running to load a user, which still feels duplicated. We also shouldn't be calling GitClient directly from the UI, sooo... feels like we should have a User class with static methods that can load itself, it we don't have a Repository/RepositoryManager to do it for us? And then everyone calls that?

shana:
Also the UI is janky but then again that's not the main point of this PR (is the screenshot up to date?) so I'm ok with merging this and polishing separately.

PR #327, opened 2017-09-18
Removing some stray meta files

StanleyGoldman:


shana:
I actually just fixed this on master :P

PR #330, opened 2017-09-19
Consolidating initializing message to a single line

AkashGutha:
### Requirements

Consolidating the logging info into a single line.

### Description of the Change

Still a work in progress. So, the idea is to implement a combined logger.
is that the aim of #203

### Alternate Designs

If not we can just remove one of the log lines

### Benefits

Having a combined logger will make things simpler going forward.

### Possible Drawbacks

I possibly don't see any.

### Applicable Issues

#203


AkashGutha:
Messed up something in the original branch. So, I had to redo the pull again.

StanleyGoldman:
Hey @AkashGutha normally we don't close and re-open pull requests to make changes. The history of what I asked you to change and why I asked you to change it is lost.

Normally, you push additional commits to the same branch or you can force push to the same branch.

src/UnityExtension/Assets/Editor/GitHub.Unity/EntryPoint.cs
```diff
@@ -34,10 +34,11 @@ private static void Initialize()
 
             // this will initialize ApplicationManager and Environment if they haven't yet
             var logPath = Environment.LogPath;
+            var logInfo = "";
```
shana: If you're only using this variable inside the block, you shouldn't declare outside the block

src/UnityExtension/Assets/Editor/GitHub.Unity/EntryPoint.cs
```diff
@@ -53,7 +54,8 @@ private static void Initialize()
                     Logging.Error(ex, "Error rotating log files");
                 }
 
-                Debug.Log("Initialized GitHub for Unity log file: " + logPath);
+                logInfo += " and Unity log file: " + logPath;
+                Debug.Log(logInfo);
```
shana: There's really no need to have a separate variable to store the message before outputting, you're overcomplicating things here.

AkashGutha: Okay, I'll just redo that

AkashGutha:
@shana can you check the changes. this must be fine.

StanleyGoldman:
Hey @AkashGutha you have to be a bit more considerate sometimes. I believe @shana is traveling.

It's okay though, I got it from here.
I tweaked the message a bit to use two lines.

img

I think it looks pretty good!
Thanks for your help.

We look forward to your next pull request.

StanleyGoldman:
img

PR #333, opened 2017-09-20
Throw if the manager isn't ready so that windows use their cached data instead

shana:
If let's say the history view is open when Unity reloads the domain, it will call Repository.Log, which might happen before the RepositoryManager is actually initialized, which would make the Log method return an empty list, and so the history view would wipe out its serialized data and show an empty list until everything else came up. We want to keep showing the serialized UI until we get information otherwise, so if there's no repository manager we really want to throw to make sure the task chain fails.


StanleyGoldman:
I notice the effect most in that I can enter/exit play mode without seeing the history view clear and repopulate.
Awesome.

PR #341, opened 2017-09-21
Reduce verbosity of build script so it's easier to read

shana:
The build.sh script was set up with a lot of verbosity to help debug, but now we don't need all of this, so reduce it to the essential output. Keep strict error checking and exit early if subcommands exit with errors flags.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -1,4 +1,4 @@
-#!/bin/sh -eux
+#!/bin/sh -eu
 Configuration="dev"
 if [ $# -gt 0 ]; then
 	Configuration=$1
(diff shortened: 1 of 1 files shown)


StanleyGoldman:
I was curious...

http://heirloom.sourceforge.net/sh/sh.1.html
```
set [--aefhknptuvx [arg ...]]
 	
-- 	No effect; useful if the first arg begins with -.
-a 	Export any variables that are modified or created from now on.
-e 	If non interactive then exit immediately if a command fails.
-f 	File name generation is disabled.
-h 	When a function is defined, look up all external commands it contains as described for the hash special command. Normally, these commands are looked up when they are executed.
-k 	All keyword arguments are placed in the environment for a command, not just those that precede the command name.
-m 	Enables job control (see below).
-n 	Read commands but do not execute them.
-p 	Makes the shell privileged. A privileged shell does not execute the system and user profiles; if an non-privileged shell (the default) has an effective user or group id different to its real user or group id or if it has an effective user or group id below 100, it resets its effective user or group id, respectively, to the corresponding real id at startup.
-t 	Exit after reading and executing one command.
-u 	Treat unset variables as an error when substituting.
-v 	Print shell input lines as they are read.
-x 	Print commands and their arguments as they are executed.
- 	Turn off the -x and -v options.

These flags can also be used upon invocation of the shell. The current set of flags may be found in $-.

If + is used instead of -, the given flags are disabled.

Remaining arguments are positional parameters and are assigned, in order, to $1, $2, etc. If no arguments are given then the values of all names are printed. 
```

shana:
That's where the flag list is! Gah couldn't find it anywhere.

PR #348, opened 2017-10-03
Added Unity DLLs path relative to Unity installation directory

mpOzelot:
Noticed that there are a lot of files called `UnityEngine.dll` and `UnityEditor.dll` (probably they are all the same). Also figuring out how pull requests work.


StanleyGoldman:
Hey @mpOzelot ... thanks for the conbtribution!

So I think you have the right idea, but maybe we should make it the best of both worlds...
I was thinking...

> copy these DLLs  in `Unity/Editor/Data/Managed` from your Unity installation into the `lib` 

What do you think?

docs/contributing/how-to-build.md
```diff
@@ -40,11 +40,11 @@ To be able to authenticate in GitHub for Unity, you'll need to:
 - [Register a new developer application](https://github.com/settings/developers) in your profile.
 - Copy [common/ApplicationInfo_Local.cs-example](../../common/ApplicationInfo_Local.cs-example) to `common/ApplicationInfo_Local.cs` and fill out the clientId/clientSecret fields for your application.
 
-The build needs to reference `UnityEngine.dll` and `UnityEditor.dll`. These DLLs are included with Unity. If you've installed Unity in the default location, the build will be able to find them automatically. If not, copy these DLLs from your Unity installation into the `lib` directory in order for the build to work.
+The build needs to reference `UnityEngine.dll` and `UnityEditor.dll`. These DLLs are included with Unity. If you've installed Unity in the default location, the build will be able to find them automatically. If not, copy these DLLs from `Unity/Editor/Data/Managed` into the `lib` directory in order for the build to work.
```
StanleyGoldman: So I think you have the right idea, but maybe we should make it the best of both worlds...
I was thinking...

> copy these DLLs  in `Unity/Editor/Data/Managed` from your Unity installation into the `lib` 

What do you think?

shana: Yeah, we should be explicit about where things are. Would
```
copy these DLLs from `[your Unity installation path]\Unity\Editor\Data\Managed`
```

make sense to users? This is really only going to affect Windows users since on mac there's no way to tell the installer to install anywhere else.

mpOzelot: Yes, what you propose is even more clear.

shana:
Thank you @mpOzelot!

Issue #248, opened 2017-08-30
Add .editorconfig to ease contribution to the project


MunchyYDL:
I think it would be a good idea to add a .editorconfig file with some of the projects code styles configured. As this would potentially minimise the amount of formatting errors in future PRs. :)

StanleyGoldman:
I agree, an `.editorconfig` is probably a good thing.

I always forget that `.editorconfig` is just the basic stuff like tabs/spaces and tab/space width and such. I wish there were something really cool for code formatting, but that only always just get's you 4/5's of the way there.

Feel free to take this on if you like.

MunchyYDL:
Sure, I will grab this and get an initial configuration in place. :)

divayprakash:
@MunchyYDL Working on it!

garg000dhruv:
In the spirit of _HacktoberFest_, I can pick this up. Let me know if you guys still need it.

idealadarsh:
@leereilly @StanleyGoldman I'd like to work on this issue.

StanleyGoldman:
Cool beans man.

Frozenfire92:
Was looking at #350 and it seems it is on hold and the repository deleted. Having some experience with editorconfig files I hope to help clear up some confusion and start some discussion.

From the [website](http://editorconfig.org):

> EditorConfig helps developers define and maintain consistent coding styles between different editors and IDEs. The EditorConfig project consists of a file format for defining coding styles and a collection of text editor plugins that enable editors to read the file format and adhere to defined styles. EditorConfig files are easily readable and they work nicely with version control systems.

An analogy from many javascript projects (not sure c# equivalents)

- EditorConfig sets the text editors tab/spaces, width, etc when working on this project
- eslint, jshint, etc are tools for validating rules (with optional editor plugins), which could also test for EditorConfig rules

From what I can tell, this [projects styleguide](https://github.com/github-for-unity/Unity/blob/master/docs/contributing/styleguide.md) most of these rules would fall into the linter category, which I think is beyond the scope of EditorConfig.

It seems many of the options in #350 come from [Visual Studio 2017's integration reference](https://docs.microsoft.com/en-us/visualstudio/ide/editorconfig-code-style-settings-reference) and potentially this [article](http://kent-boogaart.com/blog/editorconfig-reference-for-c-developers)

[...]

PR #350, opened 2017-10-08
Adds editorconfig file.

idealadarsh:
### Requirements

* Filling out the template is required. Any pull request that does not include enough information to be reviewed in a timely manner may be closed at the maintainers' discretion.
* All new code requires tests to ensure against regressions

### Description of the Change

Adds .editorconfig file.

### Benefits

New contributors will be able to contribute more easily because of minimize in code formatting error.

### Possible Drawbacks

No huge drawbacks, just the code will be formatted well.

### Applicable Issues

Issue #248

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,119 @@
+root = true
+
+# XML Project Files
+[*.{csproj,vbproj,vcxproj,vcxproj.filters,proj,projitems,shproj}]
+indent_size = 2
+
+# Configuration Files
+[*.{json,xml,yml,config,props,targets,nuspec,resx,ruleset,vsixmanifest,vsct}]
+indent_size = 2
+
+# Markdown Files
+[*.md]
+trim_trailing_whitespace = false
+
+# Web Files
+[*.{htm,html,js,ts,css,scss,less}]
+indent_size = 2
+insert_final_newline = true
+
+# Dotnet Code Style Settings
+# See https://docs.microsoft.com/en-us/visualstudio/ide/editorconfig-code-style-settings-reference
+# See http://kent-boogaart.com/blog/editorconfig-reference-for-c-developers
+[*.{cs,csx,cake,vb}]
+dotnet_sort_system_directives_first = true:warning
+dotnet_style_coalesce_expression = true:warning
+dotnet_style_collection_initializer = true:warning
+dotnet_style_explicit_tuple_names = true:warning
+dotnet_style_null_propagation = true:warning
+dotnet_style_object_initializer = true:warning
+dotnet_style_predefined_type_for_locals_parameters_members = true:warning
+dotnet_style_predefined_type_for_member_access = true:warning
+dotnet_style_qualification_for_event = true:warning
+dotnet_style_qualification_for_field = true:warning
+dotnet_style_qualification_for_method = true:warning
+dotnet_style_qualification_for_property = true:warning
+
+# Naming Symbols
+# constant_fields - Define constant fields
+dotnet_naming_symbols.constant_fields.applicable_kinds = field
+dotnet_naming_symbols.constant_fields.required_modifiers = const
+# non_private_readonly_fields - Define public, internal and protected readonly fields
+dotnet_naming_symbols.non_private_readonly_fields.applicable_accessibilities = public, internal, protected
+dotnet_naming_symbols.non_private_readonly_fields.applicable_kinds = field
+dotnet_naming_symbols.non_private_readonly_fields.required_modifiers = readonly
+# static_readonly_fields - Define static and readonly fields
+dotnet_naming_symbols.static_readonly_fields.applicable_kinds = field
+dotnet_naming_symbols.static_readonly_fields.required_modifiers = static, readonly
+# private_readonly_fields - Define private readonly fields
+dotnet_naming_symbols.private_readonly_fields.applicable_accessibilities = private
+dotnet_naming_symbols.private_readonly_fields.applicable_kinds = field
+dotnet_naming_symbols.private_readonly_fields.required_modifiers = readonly
+# public_internal_fields - Define public and internal fields
+dotnet_naming_symbols.public_internal_fields.applicable_accessibilities = public, internal
+dotnet_naming_symbols.public_internal_fields.applicable_kinds = field
+# private_protected_fields - Define private and protected fields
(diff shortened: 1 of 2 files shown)


idealadarsh:
@MunchyYDL I added some info about .editorconfig file and its usability in CONTRIBUTING.md

StanleyGoldman:
Hey @idealadarsh and @MunchyYDL 

I have to admit I'm rather new to editorconfig. I'm a big Resharper person myself.
So what IDEs/Plugins are you guys using?
Are you guys are working in Visual Studio 2015?
From what I see there is an extension in VS2015 but VS2017 applies it already?
Is there a way to format a file based on the editorconfig so I can see how it affects the file?

idealadarsh:
@StanleyGoldman EditorConfig helps developers define and maintain consistent coding styles between different editors and IDEs. No plugin is required for major IDEs. For a better understanding, go to [editorconfig.org](http://editorconfig.org)

StanleyGoldman:
@idealadarsh ...I know what it is, and actually a plugin is required for VS2015.

shana:
Hi @idealadarsh, thanks for the contribution!

I'm wondering how you generated this editorconfig file. We have some coding style guidelines in https://github.com/github-for-unity/Unity/blob/master/docs/contributing/styleguide.md, was the editorconfig based on that?

shana:
@MunchyYDL Our codebase uses spaces for indentation, so using tabs is not an option here I'm afraid. As for braces, we have them on a separate line for pretty much everything except one-liner properties and some lambdas (it really depends on readability there, if having the brace on a new line makes a lambda easier to read, then that's where it goes, and vice-versa)

.editorconfig
```diff
@@ -0,0 +1,79 @@
...
+csharp_indent_braces = false:warning
+csharp_indent_case_contents = true:warning
+csharp_indent_labels = no_change:warning
+csharp_indent_switch_labels = true:warning
+# Style Options
+csharp_style_conditional_delegate_call = true:warning
+csharp_style_expression_bodied_accessors = true:warning
+csharp_style_expression_bodied_constructors = true:warning
+csharp_style_expression_bodied_indexers = true:warning
+csharp_style_expression_bodied_methods = true:warning
```
shana: This and a lot of other entries in here are C# 6, which Unity does not support and won't compile. Have you opened our code and checked what warnings editorconfig outputs? You should make sure all entries match what Unity can compile by default (because we have a lot of code currently being compiled in Unity 5.6, so only up to C# 4).

MunchyYDL:
@shana Yeah I know, but I would prefer it to be tabs, as it gives me better options for formatting the code the way I like it to while working, as tabs can be converted in the Editor to be 2 or 4 spaces, based on your personal preferences. With spaces, it's much more constrained to what's in the file.

But, this is just my personal preference, I'm not trying to dictate what's appropriate in this codebase, I'm only trying to explain why I prefer this option. :)

idealadarsh:
@shana @MunchyYDL With all due respect I created the .editorconfig file probably with all possible formatting options to ensure everybody's need.
This is not for any specific C# version, it should work well with C# 4 as well and you can simply set the value to false or just comment out the option you don't need.
Also, it doesn't cause problem in compilation.
I just wanted to provide the best .editorconfig file. Now, I see that different people prefer different coding scheme and that is what I kept in mind when I was creating the file.
Regular contributors of this repository need to talk together for a scheme and just change the values accordingly.

StanleyGoldman:
> With all due respect I created the .editorconfig

Don't take us the wrong way... we are honestly happy you are here... 🍻 

We just had too many questions too quickly.. 😅 💦 

>Regular contributors of this repository need to talk together for a scheme and just change the values accordingly.

We agree... 👍 ... and that's why we are here 😆 

So, @idealadarsh 🎖,  help us out and lets make the necessary changes, together, here in this pull request. 

Can you try to match the `.editorconfig` to what we have in our code?

Ideally if we could hit "Reformat Code with EditorConfig™!!!" (not that such a thing existed, nor would we do it if we could, but IF 😆... IF we did) Applying this `.editorconfig` to our code should cause as little of an effect as possible.

Get what I mean?

Also, thanks for the help.

shana:
@idealadarsh As @StanleyGoldman says, we appreciate the help! It is, however, important to understand the function of an `.editorconfig` file within a project, so it will actually be useful once the PR is merged.

The function of the `.editorconfig` file is to inform contributors on what the coding styles and type of code they should be writing, so that they don't have to guess. Which is a really valuable thing that we definitely should have! More information and guidance is always better.

So when you say the following:

> This is not for any specific C# version, it should work well with C# 4 as well and you can simply set the value to false or just comment out the option you don't need.
> Also, it doesn't cause problem in compilation.

This is confusing the purpose of what the file is meant to do. Yes, it doesn't cause compilation problems: it doesn't change the existing code, therefore you don't notice it immediately. However, it will start recommending that existing code be changed to a different C# level, and if a contributor follows the recommendation, the code will not compile in Unity (but it might actually compile in VS, which will lead to even more confusion!), and this is the opposite of what the `.editorconfig` file is supposed to be doing - it's supposed to guide you in writing the code that works, not writing code that doesn't work. We can't just take an `.editorconfig` that doesn't match our codebase into our code, that's not going to work out very well.

> Now, I see that different people prefer different coding scheme and that is what I kept in mind when I was creating the file.

Personal preferences are really quite irrelevant for this, so you don't have to worry about that. What is important in the codebase is consistency, so that anyone coming in can easily read the code and find everything, which means the `.editorconfig` file should reflect what the codebase requires, not what the particular contributor likes.

Yup, I would totally prefer tabs instead of spaces too (tabs rule yo!), but given that we've already settled on spaces, we can't change it back to tabs without a huge global change to the code - it's not something you can just do without consequences. Given that tabs and spaces mixed together is even worse, these are things that should be specified in the `.editorconfig` file so people know what to use.

For reference, we already have a lot of settings specified in the https://github.com/github-for-unity/Unity/blob/master/GitHub.Unity.sln.DotSettings file (for resharper). This file performs a similar function to `.editorconfig`, so you'll probably find it useful in determining the details of the coding style for our project.

idealadarsh:
@StanleyGoldman @shana @MunchyYDL I made changes to the editorconfig file. Please review them.

idealadarsh:
Is there anybody to check the changes I made?

StanleyGoldman:
Hey @idealadarsh it's on our radar, we are working to get a new release out the door.

PR #358, opened 2017-10-11
Throwing an exception when token's username is not the expeted username

StanleyGoldman:


src/GitHub.Api/Application/ApiClient.cs
```diff
@@ -72,11 +68,10 @@ public ApiClient(UriString hostUrl, IKeychain keychain, IGitHubClient githubClie
...
         }
 
-        public async Task GetOrganizations(Action<IList<Organization>> callback)
+        public async Task GetOrganizations(Action<IList<Organization>> onSuccess, Action<Exception> onError = null)
         {
-            Guard.ArgumentNotNull(callback, "callback");
-            var organizations = await GetOrganizationInternal();
-            callback(organizations);
+            Guard.ArgumentNotNull(onSuccess, nameof(onSuccess));
+            await GetOrganizationInternal(onSuccess, onError);
```
StanleyGoldman: Why do we do this?

StanleyGoldman: Using the `...Internal` method that is...

StanleyGoldman:
Closing this because it's mostly no different than #300

PR #366, opened 2017-10-16
Preliminary implementation of git checkout

mpOzelot:
### Description of the Change
Added support for `git checkout` command

### Benefits
This functionality will be later on used for the 'discard' command in the changes window


StanleyGoldman:
Duplicate #367

Issue #356, opened 2017-10-11
Feature: Add functionality to discard changes


mpOzelot:
# Discarding changes

## Problem

When there are changes to files that the user isn't interested in keeping, they need a way of throwing away these changes. This is normally implemented via a `Discard` action for a specific file, or for all files. 

### Git overview

Git provides a `checkout` command that changes the working directory back to a previous state. This command can take:

- a commit (usually a branch name) - updates the working directory to the state of that commit
- a filename or list of files - updates the named files to the state of the commit (the current HEAD if no commit is specified)

`git checkout [file]` or `git checkout -- [files]`, will only operate on files that are already tracked by git and will not work for new files. Also, if a file has already been added to the staging area, `checkout [file]` will not affect the changes in the staging area, it will only affect changes that have not been staged yet.

![image](https://user-images.githubusercontent.com/310137/31720974-5b9b31a4-b418-11e7-89cc-d04c49f5e26b.png)

In this situation above, in order to discard changes to the different files, we need to do:

- **README.me**: `git checkout HEAD -- README.md` 
- **package.sh**: `git checkout -- package.sh` (`git checkout HEAD -- package.sh` also works)
- **a**: `git checkout` won't work, we need to manually delete the file. 

`git clean -f` throws away new files / changes to untracked files, but won't reset tracked file changes back to what they were.

`git reset --hard` throws away all changes to tracked files but doesn't touch untracked (new) files

#### Discard changes to a single file

To implement "Discard changes to a particular file", we should:

- Check whether the file is new
	- If yes, delete it
	- If no, `git checkout HEAD -- [file]`

#### Discard changes to all files

To implement "Discard all changes", there's several alternatives:

1. **clean and reset**
- Run `git clean -f` to throw away all new (untracked) files
- Run `git reset --hard` to reset all changed (tracked) files back to unchanged

2. **checkout and delete**
- Categorize all files by new/existing (untracked/tracked)
- Delete all new (untracked) files
- Run `git checkout HEAD -- [files]` on all existing (tracked) changed files

### UI/UX

TBD

### API technical details

Given that both discard and discard all requires at least two separate actions (as talked about in https://github.com/github-for-unity/Unity/issues/356), that should probably be implemented in `RepositoryManager`. What we then need is:

**`RepositoryManager.cs` (pseudocode)**:

```
DiscardAll():

   GitClient.CleanUntrackedFiles()
   GitClient.Checkout("HEAD")
```

```
[...]

PR #367, opened 2017-10-16
Enhancements/gitcheckout

mpOzelot:
Fixes: #356 
Replacement to #366


mpOzelot:
I'm not sure how tests work, think I've made a mistake at https://github.com/mpOzelot/Unity/blob/58d14bf11e10a768e1ed2ffcc00367e579e7308b/src/tests/IntegrationTests/Events/RepositoryManagerTests.cs#L524 though, should it be `GitFileStatus.Untracked` instead? Could someone have a look if the test is even correct?

PR #368, opened 2017-10-16
Split event processing and process firing

shana:
This is ensure that we can control the order of events that get fired up from the watcher regardless of the order that they happen in in the file system. Also it should make it extra easy to test things.

### DIFF
--- a/src/GitHub.Api/Events/RepositoryWatcher.cs
+++ b/src/GitHub.Api/Events/RepositoryWatcher.cs
@@ -160,27 +160,6 @@ namespace GitHub.Unity
             return processedEventCount;
         }
 
-        enum EventType
-        {
-            None,
-            ConfigChanged,
-            HeadChanged,
-            RepositoryChanged,
-            IndexChanged,
-            RemoteBranchDeleted,
-            RemoteBranchCreated,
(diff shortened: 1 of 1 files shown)


src/GitHub.Api/Events/RepositoryWatcher.cs
```diff
@@ -388,5 +443,26 @@ public void Dispose()
...
+        private enum EventType
+        {
+            None,
+            ConfigChanged,
+            HeadChanged,
+            RepositoryChanged,
+            IndexChanged,
+            RemoteBranchDeleted,
+            RemoteBranchCreated,
+            RemoteBranchChanged,
```
StanleyGoldman: It's odd...
Remote branches never change... they only seem to delete and recreate...

Maybe we should remove this enum entry and that explains that fact?

StanleyGoldman:
I sent you a pull request #369

src/GitHub.Api/Events/RepositoryWatcher.cs
shana: Hmmm, good point. Maybe if the branch gets deleted and created again, we should raise it as a change?

PR #373, opened 2017-10-17
Updating data in the branches view

shana:
Update data before OnGUI whenever the repo event happens, and OnEnable when we're not in playmode


StanleyGoldman:
I threw a log message into...

https://github.com/github-for-unity/Unity/blob/4654cae0f8d9238abbe4ca4be04c316765cc1eb9/src/UnityExtension/Assets/Editor/GitHub.Unity/UI/BranchesView.cs#L303-L306

It seems to still get called when entering Play mode..

PR #376, opened 2017-10-19
Fixing bugs in the InitProjectView

StanleyGoldman:


shana:
I took those calls out because neither view uses them, but now I notice `GitPathView` is using `Enable`. `Disable` is not being used by neither view, so no reason to add the overhead of calling those (unless you predict that they're going to be needed very soon?)

StanleyGoldman:
Okay, if they aren't being used that's fine with me.
I guess I was valuing convention over performance there.

shana:
It is a slippery slope though... if we add the override but we don't implement the call in the parent, then it won't ever be called and that might be confusing. What do you think, maybe having it there is more valuable for the programmer than not having it?

PR #378, opened 2017-10-20
Edits to the branches-view branch

StanleyGoldman:
This branch targets #151


src/UnityExtension/Assets/Editor/GitHub.Unity/UI/BranchesView.cs
```diff
@@ -586,6 +558,9 @@ public void Load(IEnumerable<ITreeData> data, string title)
 
             public Rect Render(Rect rect, Action<TreeNode> singleClick = null, Action<TreeNode> doubleClick = null)
             {
+                if (!nodes.Any())
+                    return rect;
```
shana: TODO: take this out and check `IsInitialized` before calling Render

src/UnityExtension/Assets/Editor/GitHub.Unity/UI/BranchesView.cs
```diff
@@ -518,7 +490,7 @@ private Hashtable Folders
                 }
             }
 
-            public void Load(IEnumerable<ITreeData> data, string title)
+            public void Load(IEnumerable<GitBranch> data, string title)
```
shana: Revert this change

src/UnityExtension/Assets/Editor/GitHub.Unity/UI/BranchesView.cs
```diff
@@ -304,9 +281,6 @@ private void OnButtonBarGUI()
 
         private void OnTreeGUI(Rect rect)
         {
-            if (!treeLocals.IsInitialized)
-                RefreshBranchList();
```
shana: This was ensuring render doesn't get called before initialization happens, so we need to add this check in a different way now we have the cache going

PR #387, opened 2017-10-24
Initialize Workflow

StanleyGoldman:
I merged what was in master into @donokuda's branch

img


donokuda:
Some minor tweaks but here's where I'm at with this flow:

image

I ended up removing rendering the forms on the screen. If something is missing from the initial setup, such as a git executable or a missing name and email information for git, we could use the help box to guide them to where they can set it. I feel like this will simplify this view as well as teach our users how to use our extension.

Would love some thoughts on this approach! 😄

StanleyGoldman:
Fixes #414

shana:
Feels like taking out the forms is better, it was pretty crowded in there.

PR #388, opened 2017-10-26
Removing the extra InitializeRepository call

StanleyGoldman:
`ApplicationCache` calls `DefaultEnvironment.Initialize` then `DefaultEnvironment.InitializeRepository`.
Meanwhile `DefaultEnvironment.Initiailize` calls `InitializeRepository` itself.

### DIFF
--- a/src/GitHub.Api/Platform/DefaultEnvironment.cs
+++ b/src/GitHub.Api/Platform/DefaultEnvironment.cs
@@ -42,7 +42,6 @@ namespace GitHub.Unity
             UnityAssetsPath = assetsPath;
             UnityProjectPath = assetsPath.Parent;
             UnityVersion = unityVersion;
-            InitializeRepository();
         }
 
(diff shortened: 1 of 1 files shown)


src/UnityExtension/Assets/Editor/GitHub.Unity/ApplicationCache.cs
```diff
@@ -56,7 +56,6 @@ public IEnvironment Environment
                         unityVersion = Application.unityVersion;
                     }
                     environment.Initialize(unityVersion, extensionInstallPath.ToNPath(), unityApplication.ToNPath(), unityAssetsPath.ToNPath());
-                    environment.InitializeRepository(!String.IsNullOrEmpty(repositoryPath) ? repositoryPath.ToNPath() : null);
```
shana: The call that should be removed is the one in `DefaultEnvironment` and not this one

PR #389, opened 2017-10-27
Changing the struct GitBranch to use a default constructor and fields

StanleyGoldman:
The current constructor for this struct prevents us from creating a default value for the struct itself.

https://github.com/github-for-unity/Unity/blob/2f38e6a1ae10ca804900fe19240be902095dd0f3/src/GitHub.Api/Git/GitBranch.cs#L21-L28

### DIFF
--- a/src/GitHub.Api/Git/GitBranch.cs
+++ b/src/GitHub.Api/Git/GitBranch.cs
@@ -11,12 +11,15 @@ namespace GitHub.Unity
     [Serializable]
     public struct GitBranch : ITreeData
     {
-        private string name;
-        private string tracking;
-        private bool active;
+        public static GitBranch Default = new GitBranch();
+
+        public string name;
+        public string tracking;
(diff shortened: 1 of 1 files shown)


src/GitHub.Api/Git/GitBranch.cs
```diff
@@ -2,30 +2,12 @@
...
 {
-    interface ITreeData
-    {
-        string Name { get; }
-        bool IsActive { get; }
-    }
-
     [Serializable]
-    public struct GitBranch : ITreeData
+    public struct GitBranch
```
StanleyGoldman: I realize you use this interface in #151

shana:
Just remembered we can do `public static GitBranch Default = new GitBranch { };` to leverage the default struct constructor, it **should** work well enough.

StanleyGoldman:
image

PR #392, opened 2017-10-27
Changing a nullable GitStatus operations to non nullable

StanleyGoldman:
I realized there was never a situation where `GitStatus` should be null.
We do a lot of checking if `GitStatus` is null and getting the value.
I tracked it all back to an error in how `StatusOutputProcessor` is defined.
It never returns a null `GitStatus` but is defined as `GitStatus?` once that is changed the rest of the code changes accordingly.


shana:
:+1:

PR #390, opened 2017-10-27
Adding and using the CacheManager

StanleyGoldman:
Fixes #357 
Fixes #344 
Fixes #343

Depends on:
- [x] #384 
- [x] #391 
- [x] #392
- [x] #393
- [x] #394
- [x] #396 
- [x] #397
- [x] #399
- [x] #389

PR #401, opened 2017-11-02
Restructuring events in Repository with the CacheManager

StanleyGoldman:
Targets #390 

Adding events to repository that better represent the values that are changing

### DIFF
--- a/src/GitHub.Api/Git/Repository.cs
+++ b/src/GitHub.Api/Git/Repository.cs
@@ -207,7 +207,7 @@ namespace GitHub.Unity
             var managedCache = cacheContainer.GitLogCache;
             var raiseEvent = ShouldRaiseCacheEvent(cacheUpdateEvent1, managedCache);
 
-            Logger.Trace("CheckGitLogCacheEvent Current:{0} Check:{1} Result:{2}", managedCache.LastUpdatedAt,
+            Logger.Trace("Check GitLogCache CacheUpdateEvent Current:{0} Check:{1} Result:{2}", managedCache.LastUpdatedAt,
                 cacheUpdateEvent1.UpdatedTimeString ?? "[NULL]", raiseEvent);
 
             if (raiseEvent)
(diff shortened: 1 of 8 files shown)


src/GitHub.Api/Git/Repository.cs
```diff
@@ -64,64 +58,69 @@ private void CacheContainer_OnCacheInvalidated(CacheType cacheType)
...
 
-        private void FireGitLogCacheUpdated(DateTimeOffset dateTimeOffset)
+        private void HandleRepositoryInfoCacheUpdatedEvent(CacheUpdateEvent cacheUpdateEvent)
         {
-            Logger.Trace("GitLogCacheUpdated {0}", dateTimeOffset);
-            GitLogCacheUpdated?.Invoke(new CacheUpdateEvent { UpdatedTimeString = dateTimeOffset.ToString() });
+            Logger.Trace("RepositoryInfoCache Updated {0}", cacheUpdateEvent.UpdatedTimeString);
+            CurrentBranchChanged?.Invoke(cacheUpdateEvent);
+            CurrentRemoteChanged?.Invoke(cacheUpdateEvent);
+            CurrentBranchAndRemoteChanged?.Invoke(cacheUpdateEvent);
```
shana: Do we need an extra event here, given we already raise both events right before?

src/GitHub.Api/Git/Repository.cs
```diff
@@ -64,64 +58,69 @@ private void CacheContainer_OnCacheInvalidated(CacheType cacheType)
...
 
-        private void FireRepositoryInfoCacheUpdated(DateTimeOffset dateTimeOffset)
+        private void HandleBranchCacheUpdatedEvent(CacheUpdateEvent cacheUpdateEvent)
         {
-            Logger.Trace("RepositoryInfoCacheUpdated {0}", dateTimeOffset);
-            RepositoryInfoCacheUpdated?.Invoke(new CacheUpdateEvent { UpdatedTimeString = dateTimeOffset.ToString() });
+            Logger.Trace("BranchCache Updated {0}", cacheUpdateEvent.UpdatedTimeString);
+            LocalBranchListChanged?.Invoke(cacheUpdateEvent);
+            RemoteBranchListChanged?.Invoke(cacheUpdateEvent);
+            LocalAndRemoteBranchListChanged?.Invoke(cacheUpdateEvent);
```
shana: Same here, do we need this? These events are raised from the main thread so whoever needs to listen to both events can subscribe to both and get signaled before anything else happens (like OnUI or other code), so having a combined event feels redundant.

PR #406, opened 2017-11-03
Combining RepositoryManager OnLocalBranchListUpdated & OnRemoteBranchListUpdated

StanleyGoldman:
These two events are fired one after each other when they are part of the same load operation.


shana:
I'm not sure this is the best change. The local and remote branch lists are two separate pieces of data that are updated at different times, not sure what we gain here by combining the two. What advantage does this give us over just having two separate events?

StanleyGoldman:
Makes sense

PR #410, opened 2017-11-07
Cleanup in ApplicationCache

StanleyGoldman:
As mentioned in https://github.com/github-for-unity/Unity/pull/390#issuecomment-342515942
I changed `IRemoteConfigBranchDictionary` to implement `IDictionary<string,Dictionary<string,ConfigBranch>>`

### DIFF
--- a/src/GitHub.Api/Cache/CacheInterfaces.cs
+++ b/src/GitHub.Api/Cache/CacheInterfaces.cs
@@ -62,7 +62,7 @@ namespace GitHub.Unity
 
     }
 
-    public interface IRemoteConfigBranchDictionary : IDictionary<string, IDictionary<string, ConfigBranch>>
+    public interface IRemoteConfigBranchDictionary : IDictionary<string, Dictionary<string, ConfigBranch>>
     {
 
     }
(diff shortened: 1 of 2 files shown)


StanleyGoldman:
`IRemoteConfigBranchDictionary` should have never implemented `IDictionary<string, IDictionary<string, ConfigBranch>>`. Fixing that allowed me to remove a chunk of unused code.

`RemoteConfigBranchDictionary` inherits `Dictionary<string, Dictionary<string, ConfigBranch>>` as a result it automatically implements `IDictionary<string, Dictionary<string, ConfigBranch>>` which is not the same as `IDictionary<string, IDictionary<string, ConfigBranch>>`.

shana:
Also, in general we don’t have any custom implementations of IDictionary and other list/collection interfaces, so there’s no advantage in passing interfaces around (tests can use the concrete classes just fine as well)

Unrelated to Dictionary, just fyi Unity can serialize `List<T>` so we tend to rely on concrete classes for these things.

StanleyGoldman:
`RemoteConfigBranchDictionary` has functionality that is specific to Unity (i.e. runs in Mono).
Our unit tests cannot use classes that make reference to Unity's libraries. So those classes are usually represented with interfaces and stubbed in unit tests to make other classes testable.

To aid this effort classes in `GitHub.Api` in particular don't reference the Unity libraries.

StanleyGoldman:
`IRemoteConfigBranchDictionary` lives in `GitHub.Api`

PR #411, opened 2017-11-07
Changing UpdateRepositoryInfo to ClearRepositoryInfo

StanleyGoldman:
As mentioned here https://github.com/github-for-unity/Unity/pull/390#issuecomment-342514069

https://github.com/github-for-unity/Unity/blob/88c114e5ac64496dca8b8383905192aaed143f81/src/GitHub.Api/Git/Repository.cs#L590-L629

`Name` and `CloneUrl` are already calculated lazily. The functionality in `UpdateRepositoryInfo` is a duplicate of that lazy load. So it would be best to remove `UpdateRepositoryInfo` and add `ClearRepositoryInfo` which just clears the values and then lets the lazy load kick in.

### DIFF
--- a/src/GitHub.Api/Git/Repository.cs
+++ b/src/GitHub.Api/Git/Repository.cs
@@ -462,8 +462,8 @@ namespace GitHub.Unity
 
         private void ClearRepositoryInfo()
         {
-            CloneUrl = new UriString(CurrentRemote.Value.Url);
-            Name = CloneUrl.RepositoryName;
+            CloneUrl = null;
+            Name = null;
         }
 
         private void RepositoryManager_OnLocalBranchRemoved(string name)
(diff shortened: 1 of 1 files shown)


sguthals:
So CurrentRemote is guaranteed to not be null now?

StanleyGoldman:
@sguthals good catch. I meant to clear those values instead.
As the property getter code will populate those values when null.

https://github.com/github-for-unity/Unity/blob/88c114e5ac64496dca8b8383905192aaed143f81/src/GitHub.Api/Git/Repository.cs#L590-L629

sguthals:
Can you add more of a description? Why are we doing this?

StanleyGoldman:
@sguthals Feel free to just reject the pull request next time.
I added more detail.

sguthals:
@StanleyGoldman - sounds good. This is a great description. Looks all good to me!

PR #405, opened 2017-11-02
Add test scripts and documentation

meaghanlewis:
Working on simple scripts to run the tests on both Mac and Windows. Also added a `how-to-test` doc to describe the tests and how to run them.

I haven't scripted much, so feedback is appreciated!

PR #413, opened 2017-11-07
Fixes to the test PR

StanleyGoldman:
This pull request targets #405 

- Fixed a bug in `test.cmd`
- Added a test.sh that at least works in bash for windows
  - This should work on a mac: `git clean -fxd && ./build.sh Release && ./test.sh Release`


StanleyGoldman:
Let me know if this works on a mac

PR #419, opened 2017-11-08
Refactor HookupHandlers in order accurately wrap tasks

### LINKED ISSUE #424: CommitFiles & CommitAllFiles does not stop the watcher first
Discovered through debugging

StanleyGoldman:
Fixes #424 

`HookupHandlers` was refactored to use the task system in order to accurately wrap operations performed in `RepositoryManager`.
Tasks that are exclusive or batches of exclusive operations should toggle the busy flag.
Tasks that directly cause changes to the file system should stop the watcher.

| RepositoryManager Method | Tasks Invoked   | Exclusive operations | File system changes   |
|--|--|--|--|
|  CommitAllFiles  |  GitAddTask, GitCommitTask  |  👍   |    |  
|  CommitFiles  | GitAddTask, GitCommitTask    |  👍   |    |  
|  Log |  GitLogTask  |    |    |  
|  Status|   GitStatusTask |  👍   |    |  
| Fetch|  GitFetchTask  |  👍   |    |  
|  Pull| GitPullTask  | 👍    |  👍   |  
|  Push| GitPushTask  | 👍    |    |  
|   Revert |   GitRevertTask | 👍    |  👍   |  
|  RemoteAdd| GitRemoteAddTask   | 👍    |    |  
|  RemoteRemove|   GitRemoteRemoveTask | 👍    |    |  
|  RemoteChange| GitRemoteChangeTask   | 👍    |    |  
|  SwitchBranch| GitSwitchBranchesTask   | 👍    |   👍  |  
| DeleteBranch |  GitBranchDeleteTask  |  👍   |    |  
| CreateBranch |  GitBranchCreateTask  |  👍   |    |  
|  ListLocks|  GitListLocksTask  |    |    |  
|LockFile  |   GitLockTask |  👍   |    |  
| UnlockFile | GitUnlockTask   |  👍   |    |

### DIFF
--- a/src/GitHub.Api/Git/RepositoryManager.cs
+++ b/src/GitHub.Api/Git/RepositoryManager.cs
@@ -173,20 +173,20 @@ namespace GitHub.Unity
 
         public ITask CommitAllFiles(string message, string body)
         {
-            var add = GitClient.AddAll();
-            add.OnStart += t => IsBusy = true;
-            return add
-                .Then(GitClient.Commit(message, body))
-                .Finally(() => IsBusy = false);
+            var task = GitClient.AddAll()
+                .Then(GitClient.Commit(message, body));
+
+            HookupHandlers(task, true);
+            return task;
         }
 
         public ITask CommitFiles(List<string> files, string message, string body)
         {
-            var add = GitClient.Add(files);
-            add.OnStart += t => IsBusy = true;
-            return add
-                .Then(GitClient.Commit(message, body))
-                .Finally(() => IsBusy = false);
+            var task = GitClient.Add(files)
+                .Then(GitClient.Commit(message, body));
+
+            HookupHandlers(task, true);
+            return task;
         }
 
         public ITask<List<GitLogEntry>> Log()
--- a/src/GitHub.Api/Git/RepositoryManager.cs
+++ b/src/GitHub.Api/Git/RepositoryManager.cs
@@ -177,7 +177,7 @@ namespace GitHub.Unity
                 .AddAll()
                 .Then(GitClient.Commit(message, body));
 
-            return HookupHandlers(task);
+            return HookupHandlers(task, true, true);
         }
 
         public ITask CommitFiles(List<string> files, string message, string body)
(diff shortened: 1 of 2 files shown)


shana:
@jcansdale The `ITask` infrastructure is an equivalent of the TPL on .NET with custom task schedulers for running tasks on specific threads as needed. `Then`, `ThenInUI`, methods schedule the next task as a continuation (literally `ContinueWith`) of the previous task in a specific thread scheduler, `Finally`, `FinallyInUI` methods schedule continuations regardless of final task state (success or exceptions end up always in Finally methods).

The task classes shown here are implementations of `ITask` and shell out to git to run commands.

`HookupHandlers` in this case adds extra continuations to the end of the task chain (`Then` methods always return the task that was passed in), in this case stopping the watcher and setting a busy state when the task starts, and resuming the watcher/toggling the busy state when it ends (tasks have Start and End events for this purpose.

Which actually all brings me to a question: @StanleyGoldman feels like you're wanting to stop the watcher before `Add` runs and start the watcher after `Commit` finishes? Right now it only does the handlers for the commit task, not for the whole chain...

src/GitHub.Api/Git/RepositoryManager.cs
```diff
@@ -173,20 +173,20 @@ public int WaitForEvents()
...
         {
-            var add = GitClient.AddAll();
-            add.OnStart += t => IsBusy = true;
-            return add
-                .Then(GitClient.Commit(message, body))
-                .Finally(() => IsBusy = false);
+            var task = GitClient.AddAll()
+                .Then(GitClient.Commit(message, body));
+
+            HookupHandlers(task, true);
```
shana: Hmmmm this will only hook up handlers for the `Commit` task, not the `AddAll` task (which will also fire events). Is that what we want? If we want to handle the whole chain of tasks in one go, we can hook up to the `Start` of the `AddAll` task and do a `Finally` on the `Commit` one (so that if the `AddAll` fails we can still reset the busy state and restart the watcher - if the `AddAll` fails the `Commit` will never start so handlers in `HookupHandlers` for it will never run.

StanleyGoldman:
I added a test that demonstrates what @shana was saying in her previous comment.

https://github.com/github-for-unity/Unity/blob/134558c78c8c655cc41159821f5dc6a46084bbc9/src/tests/TaskSystemIntegrationTests/Tests.cs#L111-L147

I have to fix the code.

Issue #412, opened 2017-11-07
Create a cache management solution for user data


StanleyGoldman:
#390 did not address the need to cache user data

The major problem being that this data has a different lifecycle than `Repository` as this data is needed before a `Repository` instance exists.

As a consequence of not having this solution the `InitializeView` is the second of two simultaneous calls for current user data.

StanleyGoldman:
Hmm. I don't believe git path is needed.

PR #423, opened 2017-11-09
A simpler data update method to avoid spamming GitClient

StanleyGoldman:
The `UserSettingsView` is currently requesting the user name and email from the `GitClient` inside it's display loop. This PR fixes up the logic so we can refresh at a normal rate until we have a cache for user data in #412

Depends on:
- [x] #425

PR #427, opened 2017-11-09
Utilizing a cache for user data

StanleyGoldman:
Fixes: #412 

Problems:
1. user data is not cached and queried for on demand
2. the object `User` has been used as a data object from `GitClient`

Solutions:
1. Introduce `GitUser` as the data object from `GitClient`
2. Use `User` much like `Repository` as the central front end to user data and actions
   1. `User` gets initialized with a `GitClient` and initiates a process to update it's data
   2. `User` listens to invalidation events from the `GitUserCache` and updates it's data
3. Remove vestiges of `User` and user data from `Repository` and `RepositoryManager`
4. Updating functionality in `UserSettingsView` and `InitProviewView` to use the new cache

Depends on:
- [x] #426
- [x] #434
- [x] #425 
- [x] #423
- [x] #459
- [x] #460

### DIFF
--- a/src/GitHub.Api/Git/Repository.cs
+++ b/src/GitHub.Api/Git/Repository.cs
@@ -704,7 +704,7 @@ namespace GitHub.Unity
             {
                 var dateTimeOffset = managedCache.LastUpdatedAt;
                 var updateEvent = new CacheUpdateEvent { UpdatedTimeString = dateTimeOffset.ToString() };
-                HandleGitLogCacheUpdatedEvent(updateEvent);
+                HandleUserCacheUpdatedEvent(updateEvent);
             }
         }
 
@@ -749,7 +749,7 @@ namespace GitHub.Unity
 
         private void GitUserCacheOnCacheUpdated(DateTimeOffset timeOffset)
         {
-            HandleGitLogCacheUpdatedEvent(new CacheUpdateEvent
+            HandleUserCacheUpdatedEvent(new CacheUpdateEvent
             {
                 UpdatedTimeString = timeOffset.ToString()
             });
@@ -761,7 +761,7 @@ namespace GitHub.Unity
             UpdateUserAndEmail();
         }
 
-        private void HandleGitLogCacheUpdatedEvent(CacheUpdateEvent cacheUpdateEvent)
+        private void HandleUserCacheUpdatedEvent(CacheUpdateEvent cacheUpdateEvent)
         {
             Logger.Trace("GitUserCache Updated {0}", cacheUpdateEvent.UpdatedTimeString);
             Changed?.Invoke(cacheUpdateEvent);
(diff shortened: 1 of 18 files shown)


grokys:
Just to clarify as it's not clear from the title/description: this PR should only be caching _user_ data?

StanleyGoldman:
The title of this PR could be clearer, but yes, I added access to cache objects to `GitClient`, so that `GitClient` could cache and manager user data.

StanleyGoldman:
I updated the description.

src/GitHub.Api/Git/Repository.cs
```diff
@@ -669,27 +653,129 @@ public bool IsGitHub
...
+        private IGitClient gitClient;
+
+        public event Action<CacheUpdateEvent> Changed;
+
+        public User(ICacheContainer cacheContainer)
+        {
+            this.cacheContainer = cacheContainer;
+
+            cacheContainer.GitUserCache.CacheInvalidated += GitUserCacheOnCacheInvalidated;
+            cacheContainer.GitUserCache.CacheUpdated += GitUserCacheOnCacheUpdated;
```
grokys: Will the `User` objects never need to be GCd? Adding these event handlers will mean that `cacheContainer.GitUserCache` will retain a reference to each `User` object.

sguthals:
This looks good to me @StanleyGoldman - Just take a look at @grokys question above

src/GitHub.Api/Git/Repository.cs
StanleyGoldman: `User` and `Repository` are found on `DefaultEnvironment`.

https://github.com/github-for-unity/Unity/blob/2e169082e46f436b395a5e53ccd2c2bed297251a/src/GitHub.Api/Platform/IEnvironment.cs#L30-L31

https://github.com/github-for-unity/Unity/blob/2e169082e46f436b395a5e53ccd2c2bed297251a/src/GitHub.Api/Platform/DefaultEnvironment.cs#L137-L138

There is generally only one instance of `User` and `Repository` in the system created at specific points of initialization.

https://github.com/github-for-unity/Unity/blob/2e169082e46f436b395a5e53ccd2c2bed297251a/src/GitHub.Api/Platform/DefaultEnvironment.cs#L51

https://github.com/github-for-unity/Unity/blob/2e169082e46f436b395a5e53ccd2c2bed297251a/src/GitHub.Api/Platform/DefaultEnvironment.cs#L89


These objects are only "destroyed" when the domain reloads. When the domain is reloaded these objects are being fed a `CacheContainer`. These objects store their data in file based caches specifically for this reason. It will take a moment on some other threads for the `User` and `Repository` object to get what it needs to be fully functioning (they both have `Initialize` methods for that). Until they are fully functioning, they get to use their file based cache to supply data to the user interface.

Issue #409, opened 2017-11-07
Refactor Repository Initialize method to not repeatedly call git commands


shana:
As per https://github.com/github-for-unity/Unity/pull/390/files#r148985932, the Repository Initialize method shouldn't be calling git commands every time it runs, because it's going to run every time play mode is entered, which is really not something we want to do.

It probably needs to be something that happens when someone checks for the cache and the cache requests the data because it's either 1) too old, 2) has no data, 3) initialRun is true in ApplicationCache object, or in some way check that after the first run we update data.

Maybe it makes sense for the cache objects to check that flag to see if they need to ask for new data when the views ask for data?

Maybe we should store a timestamp of when Unity first started up so the cache objects can check whether the cached data is older than data and request updated data when the views ask for data? 

Maybe some other solution makes sense here.

StanleyGoldman:
https://github.com/github-for-unity/Unity/blob/ad226ddadaa621216a96583218fdf8afcbbf1d7a/src/GitHub.Api/Git/Repository.cs#L62-L65

PR #434, opened 2017-11-13
Functionality to track "first run" in cache objects

StanleyGoldman:
As mentioned in #409

We want to start using the Invalidated events on cache objects to know when we need to re-populate data.

For some of those cache objects, we want to fire an implicit invalidated on the application's "First Run" as opposed to a domain reload. That way we know these objects need to be populated immediately.

This change adds:
- functionality to track the first run time in `ApplicationCache`
- functionality to track when `ManagedCacheBase` objects are initialized
- functionality in `ManagedCacheBase.ValidateData()` to verify if cache objects were initialized after application first run time, otherwise it will update and fire an "Invalidated" event
- functionality to define which cache objects are invalid on first tun

This provides base functionality to:
- #427 
- #436 

Depends on:
- [x] #435

### DIFF
--- a/src/UnityExtension/Assets/Editor/GitHub.Unity/ApplicationCache.cs
+++ b/src/UnityExtension/Assets/Editor/GitHub.Unity/ApplicationCache.cs
@@ -19,18 +19,7 @@ namespace GitHub.Unity
         {
             get
             {
-                if (!firstRunValue.HasValue)
-                {
-                    firstRunValue = firstRun;
-                }
-
-                if (firstRun)
-                {
-                    firstRun = false;
-                    FirstRunAt = DateTimeOffset.Now;
-                    Save(true);
-                }
-
+                EnsureFirstRun();
                 return firstRunValue.Value;
             }
         }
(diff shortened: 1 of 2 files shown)


src/UnityExtension/Assets/Editor/GitHub.Unity/ApplicationCache.cs
```diff
@@ -178,6 +213,24 @@ public DateTimeOffset LastVerifiedAt
...
+                if (!initializedAtValue.HasValue)
+                {
+                    initializedAtValue = DateTimeOffset.Parse(InitializedAtString);
+                }
+
+                return initializedAtValue.Value;
+            }
+            set
+            {
+                InitializedAtString = value.ToString();
```
shana: All of these `DateTimeOffset.ToString()` calls need to specify the ISO 8016 format, eg. `2017-11-16T20:24:46+00:00`. As it is, they're outputting strings in whatever local configuration the system has, which may not be parseable by `Parse`.

StanleyGoldman: Originally fixed in #457 carried over here

StanleyGoldman:
@jcansdale Unity cannot serialize `DateTimeOffset` objects, but it will serialize `string` objects. So for all these dates we have a `string` field and a `DateTimeOffset?` field; the latter of which we populate lazily. I did however make a change based on your comment, there was no reason to set the lazily generated value to null and thus wait for the lazy generation.

sguthals:
Seems straight forward and like a clean approach.

PR #433, opened 2017-11-13
Rolling back of recent patches

StanleyGoldman:
**Note:** This pull request does not target master and will be part of a larger pull request

In order to temporarily solve the issues of #421 two patches, #431 and #428 were made.
These patches forcibly call methods to refresh sets of data that we know get corrupted by the current methods that process events. The first step towards resolving this issue is rolling back the patches.

PR #436, opened 2017-11-13
Repository using cache invalidation events to update data

StanleyGoldman:
**Note:** This branch targets `enhancements/repository-watcher-refactor-rollup` #461

Related to: #409

Depends on:
- [x] #426
- [x] #434
- [x] #433 

A simpler part of the solution to #409.

Due to the changes in #434, the git log, git lock and git status caches throw an "Invalid" event on "First Run". _(The true in the constructor..)_

https://github.com/github-for-unity/Unity/blob/8a9554ff6fb477c243bb1e4dd93932c90d9d39f5/src/UnityExtension/Assets/Editor/GitHub.Unity/ApplicationCache.cs#L884

https://github.com/github-for-unity/Unity/blob/8a9554ff6fb477c243bb1e4dd93932c90d9d39f5/src/UnityExtension/Assets/Editor/GitHub.Unity/ApplicationCache.cs#L825

https://github.com/github-for-unity/Unity/blob/8a9554ff6fb477c243bb1e4dd93932c90d9d39f5/src/UnityExtension/Assets/Editor/GitHub.Unity/ApplicationCache.cs#L766

We can use that invalid event to accurately update data in `Repository`


src/UnityExtension/Assets/Editor/GitHub.Unity/ApplicationCache.cs
```diff
@@ -10,25 +10,46 @@ namespace GitHub.Unity
...
+            }
+        }
+
+        public DateTimeOffset FirstRunAt
+        {
+            get
+            {
+                if (!firstRunAtValue.HasValue)
+                {
+                    firstRunAtValue = DateTimeOffset.Parse(firstRunAtString);
```
grokys: Should you be passing a `CultureInfo` to these parse methods, or is it OK that they're being parsed using the `CurrentCulture`?

StanleyGoldman: I don't have a good answer here, I'm going to use a lifeline (i.e. @shana)

StanleyGoldman: https://github.com/github-for-unity/Unity/issues/448

Issue #447, opened 2017-11-15
The UTC representation falls outside the 1-9999 year range


StanleyGoldman:
```
ArgumentOutOfRangeException: Argument is out of range.
Parameter name: The UTC representation falls outside the 1-9999 year range
System.DateTimeOffset.Parse (System.String input, IFormatProvider formatProvider, DateTimeStyles styles) (at /Users/builduser/buildslave/mono/build/mcs/class/corlib/System/DateTimeOffset.cs:305)
System.DateTimeOffset.Parse (System.String input, IFormatProvider formatProvider) (at /Users/builduser/buildslave/mono/build/mcs/class/corlib/System/DateTimeOffset.cs:290)
System.DateTimeOffset.Parse (System.String input) (at /Users/builduser/buildslave/mono/build/mcs/class/corlib/System/DateTimeOffset.cs:285)
GitHub.Unity.ManagedCacheBase`1[T].get_LastUpdatedAt () (at D:/code/github/Unity/src/UnityExtension/Assets/Editor/GitHub.Unity/ApplicationCache.cs:151)
GitHub.Unity.GitLocksCache.GitHub.Unity.IManagedCache.get_LastUpdatedAt ()
GitHub.Unity.Repository.ShouldRaiseCacheEvent (CacheUpdateEvent cacheUpdateEvent, IManagedCache managedCache) (at D:/code/github/Unity/src/GitHub.Api/Git/Repository.cs:277)
GitHub.Unity.Repository.CheckLocksChangedEvent (CacheUpdateEvent cacheUpdateEvent) (at D:/code/github/Unity/src/GitHub.Api/Git/Repository.cs:192)
GitHub.Unity.SettingsView.OnEnable () (at D:/code/github/Unity/src/UnityExtension/Assets/Editor/GitHub.Unity/UI/SettingsView.cs:57)
GitHub.Unity.Window.OnEnable () (at D:/code/github/Unity/src/UnityExtension/Assets/Editor/GitHub.Unity/UI/Window.cs:102)
```

Dates and times in the cache are not being serialized properly, users will happen upon this error in the new build rather soon.

PR #457, opened 2017-11-20
Serializing DateTimeOffset objects with ISO8601

StanleyGoldman:
Fixes: #447

Cache objects were being serialized with their date times just being `ToString`d.
This works fine most of the time, but of course it is not consistent.


sguthals:
You also removed using Octokit in this PR. Any reason for that?

StanleyGoldman:
@jcansdale That does not enforce the ISO8601 format. From what I understand if you do not set a format you end up with your system global format. Which in the case of some users, like @shana, who prefer a 24-Hour time format, this leads to an output that is not among the formats that are parsed for by default.

StanleyGoldman:
@sguthals It was not intentional. I think Resharper removed it, but the namespace was not used so it's a null change.

sguthals:
Looks good!

PR #460, opened 2017-11-21
Changing to a format string that is compatible with parse operations in Mono

StanleyGoldman:
In recently merged pull request #457, we used the format string `o` to format `DateTimeOffset` objects to the Iso8601 format.

Although this works for output, it does not work for parsing. Usingn `o` in `DateTimeOffset.Parse()` or `DateTimeOffset.ParseExact()` throws an exception with `Invalid Format String` as the message.

Using the format string `yyyy-MM-ddTHH\\:mm\\:ss.fffffffzzz` outputs the same content as `o`, but also has the added benefit of working in Mono/Unity.

### DIFF
--- a/src/GitHub.Api/Helpers/Constants.cs
+++ b/src/GitHub.Api/Helpers/Constants.cs
@@ -9,7 +9,7 @@ namespace GitHub.Unity
         public const string UsageFile = "usage.json";
         public const string GitInstallPathKey = "GitInstallPath";
         public const string TraceLoggingKey = "EnableTraceLogging";
-        public const string Iso8601Format = "o";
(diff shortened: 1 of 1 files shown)


sguthals:
Looks good!

grokys:
What do you mean by:

> Although this works for output, it does not work for parsing.

What doesn't work for parsing? You can't parse ISO8601 format dates on mono?

StanleyGoldman:
I never heard of the `DateTimeOffset` format string `o` until recently.
When I use that same format string in `DateTimeOffset.Parse()` or `DateTimeOffset.ParseExact()` I get an `Invalid Format String` error.

I'll update the body above to be clearer...

Issue #446, opened 2017-11-15
Update AppVeyor scripts to capture failed test logs


StanleyGoldman:
https://www.appveyor.com/docs/packaging-artifacts/#pushing-artifacts-from-scripts

https://www.appveyor.com/docs/running-tests/

https://www.appveyor.com/docs/packaging-artifacts/

PR #463, opened 2017-11-21
Changes to appveyor script to capture tests of failed builds

StanleyGoldman:
Fixes: #446 

In AppVeyor, with our build we are capturing the build artifacts and test logs, but if your tests fail the `artifacts` task does not run.

I'm debugging a set of integration tests that consistently fail on AppVeyor but work locally.
Our tests write their logging output to a file and I would like to get those outputs in AppVeyor.

The only way to do is is to script the build failure event to upload any files if found

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -57,9 +57,10 @@ test:
     except:
     - DoNotRunOnAppVeyor
     - TimeSensitive
-after_test:
-  - ps: Get-ChildItem build\*.log | % { Push-AppveyorArtifact $_.FullName -FileName $_.Name }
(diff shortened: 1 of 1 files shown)


StanleyGoldman:
You can see the result of this pull request in action here: https://ci.appveyor.com/project/github-windows/unity/build/2858.enhancements/repository-watcher-refactor/artifacts

grokys:
Out of interest, why are you doing this? Is the appveyor output not readable enough?

StanleyGoldman:
Updated the body of this pull request

PR #469, opened 2017-11-27
Functionality to mark folder nodes as "root"

### LINKED ISSUE #454: TreeControl: Globe icon being shown on folders in the local branch list
Only the remote branch list should show a globe for the top level folder nodes (which represent the remotes), but apparently the local branch is also showing it instead of using the normal folder icon. Probably bad copy paste

image

StanleyGoldman:
**Note:** This branch targets `enhancements/branches-view-rollup` #464 

Fixes #454

- Renaming icons to be something more useful
- Adding functionality to render the globe icon at the right time

Depends on:
- [x] #151 
- [x] #465
- [x] #479

### DIFF
--- a/src/UnityExtension/Assets/Editor/GitHub.Unity/UI/BranchesView.cs
+++ b/src/UnityExtension/Assets/Editor/GitHub.Unity/UI/BranchesView.cs
@@ -141,7 +141,9 @@ namespace GitHub.Unity
             if (treeLocals == null)
             {
                 treeLocals = new BranchesTree();
+
                 treeRemotes = new BranchesTree();
+                treeRemotes.IsRemote = true;
 
                 UpdateTreeIcons();
             }
@@ -167,10 +169,10 @@ namespace GitHub.Unity
                     treeLocals.ActiveNodeIcon = Styles.ActiveBranchIcon;
                 }
 
-                if (treeLocals.NodeIcon == null)
+                if (treeLocals.BranchIcon == null)
                 {
                     localsLoaded = true;
-                    treeLocals.NodeIcon = Styles.BranchIcon;
+                    treeLocals.BranchIcon = Styles.BranchIcon;
                 }
 
                 if (treeLocals.FolderIcon == null)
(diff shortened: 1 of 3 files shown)


src/UnityExtension/Assets/Editor/GitHub.Unity/UI/BranchesView.cs
```diff
@@ -134,19 +134,23 @@ private void Render()
...
         private void BuildTree()
         {
+            if (treeLocals == null)
+            {
+                treeLocals = new Tree();
+                treeLocals.ActiveNodeIcon = Styles.ActiveBranchIcon;
+                treeLocals.NodeIcon = Styles.BranchIcon;
+                treeLocals.FolderIcon = Styles.FolderIcon;
+
+                treeRemotes = new Tree();
```
sguthals: treeLocals and treeRemotes are always either defined or null together right? It isn't worth having a separate if for treeRemotes?

src/UnityExtension/Assets/Editor/GitHub.Unity/UI/TreeControl.cs
```diff
@@ -383,7 +385,7 @@ private void ResetNodeIcons(TreeNode node)
             }
             else if (node.IsFolder)
             {
-                if (node.Level == 1)
+                if (node.Level == RootFolderIconLevel)
```
sguthals: Why was this switched from 1 to -1?

src/UnityExtension/Assets/Editor/GitHub.Unity/UI/BranchesView.cs
StanleyGoldman: They always get instantiated together, so I figured I could save the `boolean` check.

src/UnityExtension/Assets/Editor/GitHub.Unity/UI/TreeControl.cs
StanleyGoldman: You know what. It's a rudimentary fix for something that i'm already having to change.
I'm going to send a PR before this one that provides a correct solution instead.

sguthals: Sounds good

PR #475, opened 2017-11-28
ChangesView Tree Implementation

StanleyGoldman:

- Added `GitStatusEntryTreeData` an object to track a `GitStatusEntry` object in a `Tree`
- Functionality to check/uncheck all nodes
- Some code cleanup in `BranchesView`
- `ChangesTreeControl` implementation and utilization in `ChangesView`
- Updates to the layout mechanics in `Tree`

Depends on:
- [x] #476
- [x] #464


src/GitHub.Api/UI/TreeBase.cs
```diff
@@ -102,6 +102,28 @@ public void Load(IEnumerable<TData> treeDatas)
             }
         }
 
+        public void CheckAll(bool isChecked)
```
sguthals: Wait - this looks like it is just toggling each node, not checking each node?

StanleyGoldman: If you send `true` it will check all nodes. If you send `false` it will uncheck all nodes.
Is it the name?
Maybe the method could be called `SetCheckOnAll`?

sguthals: The name might be best changed, but I was referring to inside the foreach.

Just looking at the code walking through it I think I'm reading it right:
110: wasChecked is true if the node is checked right now and false otherwise
111: change the node...

Ah...never mind. I see it now. Sorry - this is good to go! If you want to change the name it might be clearer, but the logic is sound.

PR #487, opened 2017-11-30
Updating ahead behind status efficiently

### LINKED ISSUE #385: Using git status to get ahead/behind values can be slow
Consider using a different command.
`git rev-list --left-right --count {local}...{remote}`
Outputs: `1 0` for ahead 1 behind 0.

### LINKED ISSUE #484: Update history view after commit is pushed
### Description
After pushing new commits to a branch, the history view does not update.

### Steps to Reproduce
After already initializing, publishing and pushing the initial commit...
- make some changes
- commit
- push changes
- wait for push confirmation

**Expected behavior:** 
The circle to the left of the commit name is filled in and the **push** button in the upper right hand corner still shows there is nothing to push

**Actual behavior:**
The commit visually appears it was not pushed because the circle to the left of the commit name is not filled in and the **push** button in the upper right hand corner still shows the number of commits to push

**Reproduces how often:** 
100%

### Build:
`github-for-unity-0.23.0.3170-a66bd286.unitypackage` from
https://github.com/github-for-unity/Unity/releases/tag/untagged-128514e09193583decb6

StanleyGoldman:
Fixes #484 
Fixes #385

A regular `git status` can get us the ahead/behind status, but it's also not the fastest way to get that information. We can instead get that information with a `git rev-list` command.

- Split up the storage of `GitStatus` in order to update entries individually
- Implement a means to get the ahead/behind status
- Integrate that functionality into `RepositoryManager`

Depends on:
- [x] #491


meaghanlewis:
LGTM!

src/tests/IntegrationTests/Events/RepositoryManagerTests.cs
```diff
@@ -162,6 +174,7 @@ await RepositoryManager
 
                 repositoryManagerListener.Received().OnIsBusyChanged(Args.Bool);
                 repositoryManagerListener.DidNotReceive().CurrentBranchUpdated(Args.NullableConfigBranch, Args.NullableConfigRemote);
+                repositoryManagerListener.DidNotReceive().GitAheadBehindStatusUpdated(Args.GitAheadBehindStatus);
```
StanleyGoldman: Commit operations will affect the ahead/behind status. We should update this here.

src/tests/IntegrationTests/Events/RepositoryManagerTests.cs
```diff
@@ -211,6 +227,7 @@ await RepositoryManager
...
```
StanleyGoldman: Same thing here

src/tests/IntegrationTests/Events/RepositoryManagerTests.cs
```diff
@@ -234,6 +251,7 @@ await RepositoryManager
...
```
StanleyGoldman: Same thing here.

src/tests/IntegrationTests/Events/RepositoryManagerTests.cs
```diff
@@ -435,6 +469,7 @@ await RepositoryManager
 
                 repositoryManagerListener.Received().OnIsBusyChanged(Args.Bool);
                 repositoryManagerListener.Received().CurrentBranchUpdated(Args.NullableConfigBranch, Args.NullableConfigRemote);
+                repositoryManagerListener.DidNotReceive().GitAheadBehindStatusUpdated(Args.GitAheadBehindStatus);
```
StanleyGoldman: Removing a remote might affect the remote for the current branch. We should be triggering this here.

src/tests/IntegrationTests/Events/RepositoryManagerTests.cs
```diff
@@ -457,6 +492,7 @@ await RepositoryManager
...
```
StanleyGoldman: Adding a remote might affect the remote for the current branch. We should be triggering this here.

src/tests/IntegrationTests/Events/RepositoryManagerTests.cs
```diff
@@ -526,6 +567,7 @@ await RepositoryManager.SwitchBranch("branch2")
...
```
StanleyGoldman: When switching branches we should probably see a `GitStatusUpdated` event.

src/tests/IntegrationTests/Events/RepositoryManagerTests.cs
```diff
@@ -569,6 +614,7 @@ await RepositoryManager.SwitchBranch("branch2")
 
                 repositoryManagerListener.Received().OnIsBusyChanged(Args.Bool);
                 repositoryManagerListener.DidNotReceive().CurrentBranchUpdated(Args.NullableConfigBranch, Args.NullableConfigRemote);
+                repositoryManagerListener.DidNotReceive().GitAheadBehindStatusUpdated(Args.GitAheadBehindStatus);
```
StanleyGoldman: A pull operation should probably trigger something.

PR #488, opened 2017-11-30
Updating LFS locks list after lock operation

### LINKED ISSUE #480: Show visual indication that files are locked
### Description
Files can be locked but there is no visual indication that the file has been successfully locked.

### Steps to Reproduce
Go to request a lock on a file

**Expected behavior:** 
A lock icon to appear on the locked file and for the file to appear under Settings in the Locked files section

**Actual behavior:**
The file doesn't appear to be locked and there is no visual indication that it is

**Reproduces how often:** 
100% on mac and windows

### Build:
`github-for-unity-0.23.0.3170-a66bd286.unitypackage` from
https://github.com/github-for-unity/Unity/releases/tag/untagged-128514e09193583decb6

StanleyGoldman:
Fixes #480

- Added hook for locks updated event
- Added check for current remote url before checking for locks
- Added explicit check for locks after lock operation
- Added a missing null check


meaghanlewis:
This LGTM

PR #489, opened 2017-11-30
Removing forced update of config data on a mac

StanleyGoldman:
These lines of code should not be needed, it's explicit to mac and forcing an re-read of config data.
But there is already a watcher on the config file that will cause a re-read on the data.


meaghanlewis:
I haven't noticed any difference in behavior with this change, so LGTM 👍

PR #491, opened 2017-12-01
Making integration tests more solid

StanleyGoldman:
Playing with tests in another branch. I realized some inconsistencies in our tests.

1.  We were not being sure to call Reset on the `AutoResetEvent`, which was causing us some troubles.
2. There is an initial log event that fires upon startup, we should be capturing that so it doesn't interfere with tests

In these tests we should be waiting for events we expect to wait for what events we expect to receive.

### DIFF
--- a/src/tests/IntegrationTests/Events/RepositoryManagerTests.cs
+++ b/src/tests/IntegrationTests/Events/RepositoryManagerTests.cs
@@ -230,12 +230,13 @@ namespace IntegrationTests
 
                 repositoryManagerEvents.GitStatusUpdated.WaitOne(Timeout).Should().BeTrue();
                 repositoryManagerEvents.LocalBranchesUpdated.WaitOne(Timeout).Should().BeTrue();
+                repositoryManagerEvents.GitLogUpdated.WaitOne(Timeout).Should().BeTrue();
 
                 repositoryManagerListener.Received().OnIsBusyChanged(Args.Bool);
                 repositoryManagerListener.DidNotReceive().CurrentBranchUpdated(Args.NullableConfigBranch, Args.NullableConfigRemote);
                 repositoryManagerListener.Received().GitStatusUpdated(Args.GitStatus);
                 repositoryManagerListener.DidNotReceive().GitLocksUpdated(Args.GitLocks);
-                repositoryManagerListener.DidNotReceive().GitLogUpdated(Args.GitLogs);
+                repositoryManagerListener.Received().GitLogUpdated(Args.GitLogs);
                 repositoryManagerListener.Received().LocalBranchesUpdated(Args.LocalBranchDictionary);
                 repositoryManagerListener.DidNotReceive().RemoteBranchesUpdated(Args.RemoteDictionary, Args.RemoteBranchDictionary);
             }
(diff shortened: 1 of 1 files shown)


src/tests/IntegrationTests/Events/RepositoryManagerTests.cs
```diff
@@ -571,20 +585,21 @@ await RepositoryManager.SwitchBranch("branch2")
...
                 await RepositoryManager.Fetch("origin").StartAsAsync();
                 await TaskManager.Wait();
 
                 RepositoryManager.WaitForEvents();
                 repositoryManagerEvents.WaitForNotBusy();
 
-                repositoryManagerEvents.LocalBranchesUpdated.WaitOne(TimeSpan.FromSeconds(2)).Should().BeTrue();
-                repositoryManagerEvents.RemoteBranchesUpdated.WaitOne(TimeSpan.FromSeconds(2)).Should().BeTrue();
+                repositoryManagerEvents.GitLogUpdated.WaitOne(Timeout).Should().BeTrue();
+                repositoryManagerEvents.LocalBranchesUpdated.WaitOne(Timeout).Should().BeTrue();
```
sguthals: Why do we not longer need a timeout on RemoteBranchesUpdated?

src/tests/IntegrationTests/Events/RepositoryManagerTests.cs
```diff
@@ -82,7 +84,8 @@ public override void OnSetup()
                 RepositoryManager.WaitForEvents();
                 repositoryManagerEvents.WaitForNotBusy();
 
-                repositoryManagerEvents.GitStatusUpdated.WaitOne(TimeSpan.FromSeconds(20)).Should().BeTrue();
+                repositoryManagerEvents.GitStatusUpdated.WaitOne(Timeout).Should().BeTrue();
+                repositoryManagerEvents.GitLogUpdated.WaitOne(Timeout).Should().BeTrue();
```
sguthals: GitLogUpdate timeout was also missing originally?

StanleyGoldman:
@sguthals the trick is to wait for the events you expect to be fired. In some cases we were getting events when we didn't expect or later than we expected. This pull request is cleaning all that up.

sguthals:
So mainly just cleaning up what should have been waiting and what shouldn't have - just making sure those were all explicit :)

src/tests/IntegrationTests/Events/RepositoryManagerTests.cs
```diff
@@ -541,13 +565,13 @@ await RepositoryManager.SwitchBranch("branch2")
...
 
-                repositoryManagerEvents.LocalBranchesUpdated.WaitOne(TimeSpan.FromSeconds(2)).Should().BeTrue();
+                repositoryManagerEvents.LocalBranchesUpdated.WaitOne(Timeout).Should().BeTrue();
 
                 repositoryManagerListener.Received().OnIsBusyChanged(Args.Bool);
                 repositoryManagerListener.DidNotReceive().CurrentBranchUpdated(Args.NullableConfigBranch, Args.NullableConfigRemote);
                 repositoryManagerListener.DidNotReceive().GitStatusUpdated(Args.GitStatus);
                 repositoryManagerListener.DidNotReceive().GitLocksUpdated(Args.GitLocks);
-                repositoryManagerListener.Received().GitLogUpdated(Args.GitLogs);
+                repositoryManagerListener.DidNotReceive().GitLogUpdated(Args.GitLogs);
```
StanleyGoldman: This surprised me, we should be updating the log after a pull operation.

StanleyGoldman: @meaghanlewis After we get a new build going you should be able to reproduce a bug where pulling from a repository does not update the log.

StanleyGoldman: We can follow this up with a future pull request.

PR #501, opened 2017-12-04
More fixes revolving around DateTimeOffset serialization

StanleyGoldman:
- CacheUpdateEvent does not deserailize itself properly
- Application cache objects have their initial time strings set improperly


src/GitHub.Api/Git/Repository.cs
```diff
@@ -726,15 +726,23 @@ private void UpdateUserAndEmail()
     public struct CacheUpdateEvent
     {
         [NonSerialized] private DateTimeOffset? updatedTimeValue;
-        private string updatedTimeString;
+        public string updatedTimeString;
```
sguthals: Is this variable still needed? Seems like it was replaces with UpdatedTime?

StanleyGoldman: Unity cannot serialize `DateTime` and `DateTimeOffset` objects.
We are using `string` as a backing store because this is what Unity serializes when the domain reloads.
It is our "hard copy" so to speak.

After the domain reloads `updatedTimeValue` will be null and it will be repopulated using `updatedTimeString`.

PR #511, opened 2017-12-06
Toggling checked state in Tree view

StanleyGoldman:

Functionality to toggle checked state in `Tree`.
To properly handle checks and unchecks we have two major tasks:
- Ripple checks on folders down to all children recursively
- Ripple checks on items up through their folders recursively

What makes this tricky is that `nodes` is a **flat list**. It is not an object with a Children property that can be traversed.

Also added functionality to perform a check on spacebar.

### DIFF
--- a/src/UnityExtension/Assets/Editor/GitHub.Unity/UI/TreeControl.cs
+++ b/src/UnityExtension/Assets/Editor/GitHub.Unity/UI/TreeControl.cs
@@ -30,6 +30,7 @@ namespace GitHub.Unity
         [SerializeField] private List<TreeNode> nodes = new List<TreeNode>();
         [SerializeField] private TreeNode selectedNode = null;
         [SerializeField] private TreeNodeDictionary folders = new TreeNodeDictionary();
+        [SerializeField] private TreeNodeDictionary checkedFileNodes = new TreeNodeDictionary();
 
         [NonSerialized] private Stack<bool> indents = new Stack<bool>();
         [NonSerialized] private Action<TreeNode> rightClickNextRender;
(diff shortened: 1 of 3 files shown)


src/UnityExtension/Assets/Editor/GitHub.Unity/UI/TreeControl.cs
```diff
@@ -257,26 +255,116 @@ public void Blur()
...
+                    ToggleChildrenChecked(i, childNode, isChecked);
                 }
+            }
+        }
+
+        private void ToggleParentFoldersChecked(int idx, TreeNode node, bool isChecked)
+        {
+            while (true)
+            {
+                if (node.Level > 0)
```
sguthals: If node.level > 0 we are also guaranteeing that idx > 0? Otherwise i = idx - 1 could through out of bounds?

src/UnityExtension/Assets/Editor/GitHub.Unity/UI/TreeControl.cs
```diff
@@ -257,26 +255,116 @@ public void Blur()
...
+                            if (isChecked != previousNodeIsChecked)
+                            {
+                                siblingsInSameState = false;
+                            }
+                        }
+                    }
+
+                    if (siblingsInSameState)
+                    {
+                        for (var i = idx + 1; i < nodes.Count && node.Level <= nodes[i].Level; i++)
```
sguthals: Same here for not being the last node? Guaranteed?

src/UnityExtension/Assets/Editor/GitHub.Unity/UI/TreeControl.cs
```diff
@@ -257,26 +255,116 @@ public void Blur()
...
+                            var followingNodeIsChecked = followingNode.CheckState == CheckState.Checked;
+                            if (isChecked != followingNodeIsChecked)
+                            {
+                                siblingsInSameState = false;
+                                break;
+                            }
+                        }
+                    }
+
+                    var parentIndex = firstSiblingIndex - 1;
```
sguthals: Same here for firstSiblingIndex not being 0?

sguthals:
Those final changes that you added after my questions look good.

sguthals:
Removing files from checked list LGTM

src/UnityExtension/Assets/Editor/GitHub.Unity/UI/TreeControl.cs
StanleyGoldman: Levels run from -1 and up. So if level > 0 then idx must be > 0.

StanleyGoldman: We are also going backwards from our current position. So we don't make any assumptions about the structure, and the for loop bounds will keep us from checking idx of -1 anyway.

src/UnityExtension/Assets/Editor/GitHub.Unity/UI/TreeControl.cs
StanleyGoldman: Here we are going forwards from our current position, and the for loop bounds will keep us in the list. We are looking for the end of the list or a point where the level is higher than the current position. We should be good here.

src/UnityExtension/Assets/Editor/GitHub.Unity/UI/TreeControl.cs
StanleyGoldman: We would not have entered this routine if the level was not at least 1.
So the first index this will ever work on is 1. Meaning our we can safely expect to find a parent node at 0.

PR #513, opened 2017-12-07
Refactor TreeView

StanleyGoldman:

Initially tree loading functionality was isolated to `GitHub.Api` in order to be able to be tested.
This is not the only functionality around `Tree` that needs to be testable.
The functionality to toggle visibility and checked state is also quite complex and could serve to be isolated.
Also with some crafty use of generics we can make the additional tree types easier to implement.

- Created `TreeBase<TNode, TData>`
  - Changed `Tree` to `Tree<TNode, TData>: TreeBase<TNode, TData>`
  - Moved `Load` from `TreeLoader`
    - Deleted `ITree`
  - Moved `ToggleNodeVisibility`, `ToggleNodeChecked` from `Tree` to `TreeBase`

Additionally added an `IconBadge` to `TreeNode` and it's `Render`


src/GitHub.Api/UI/TreeBase.cs
```diff
@@ -0,0 +1,302 @@
...
+            var nodeLevel = node.Level;
+            node.IsCollapsed = !node.IsCollapsed;
+            idx++;
+            for (; idx < Nodes.Count && Nodes[idx].Level > nodeLevel; idx++)
+            {
+                Nodes[idx].IsHidden = node.IsCollapsed;
+                if (Nodes[idx].IsFolder && !node.IsCollapsed && Nodes[idx].IsCollapsed)
+                {
+                    var level = Nodes[idx].Level;
+                    for (idx++; idx < Nodes.Count && Nodes[idx].Level > level; idx++)
```
sguthals: Is this just to get to the last node?

src/GitHub.Api/UI/TreeBase.cs
```diff
@@ -0,0 +1,302 @@
...
+
+            switch (node.CheckState)
+            {
+                case CheckState.Mixed:
+                case CheckState.Empty:
+                    node.CheckState = CheckState.Checked;
+                    isChecked = true;
+                    break;
+
+                case CheckState.Checked:
```
sguthals: You don't need to `isChecked = false;` here?

src/GitHub.Api/UI/TreeBase.cs
StanleyGoldman: Hmm. This is @shana's logic which i'm moving from place to place.

src/GitHub.Api/UI/TreeBase.cs
StanleyGoldman: It is set to false by default a few lines up.

src/GitHub.Api/UI/TreeBase.cs
StanleyGoldman: lol... uhmm yeaa... she is going to have to document that...

StanleyGoldman: We can add tests for it now.. 🐑

src/GitHub.Api/UI/TreeBase.cs
sguthals: Duh..

PR #529, opened 2017-12-13
Setting Tree properites OnEnable

StanleyGoldman:


StanleyGoldman:
Replaced with https://github.com/github-for-unity/Unity/pull/525

PR #534, opened 2017-12-13
Setting the field values on Tree objects in OnEnable

StanleyGoldman:
I'm not entirely sure why, but the values of Tree seem to get unset
randomly on domain reload. This at least helps to make sure I dont lose
these values, until I figure out what I'm doing wrong.


StanleyGoldman:
Figured it out

PR #545, opened 2017-12-19
Download Functionality

StanleyGoldman:
**Note:** This pull request targets `enhancements/async-git-setup-rollup`

- Created `DownloadTask`
  - A task that downloads a file with functionality to support resume
  - If specified the task will also verify the MD5 of a file
  - If specified the task will perform retries if there was an issue with the download
- Created `DownloadTestTask`
  - A task that downloads text from a url


src/GitHub.Api/Tasks/DownloadTask.cs
```diff
@@ -0,0 +1,316 @@
...
+
+                if (trackProgress)
+                    watch.Stop();
+
+                totalRead += bytesRead;
+
+                if (bytesRead > 0)
+                {
+                    destination.Write(buffer, 0, bytesRead);
+                    if (trackProgress)
```
sguthals: This if-statement is just figuring out how long it took to download, correct?

StanleyGoldman: That code is attempting to figure out the progress of the download in real time.
But my tests of the progress reporting has shown me it's less accurate that one would like.
I'm not using it now, but I will record an issue to address it because I would like to use it soon.

PR #548, opened 2017-12-29
Blocking the main thread and displaying a progress bar during Pull

StanleyGoldman:
From experimenting with the sample here in the documentation [EditorUtility.DisplayProgressBar](https://docs.unity3d.com/ScriptReference/EditorUtility.DisplayProgressBar.html). The progress bar on it's own does not enforce modality, the only way to do that is to actually block on the UI thread until the process is complete.

Example of non modality. This is the sample from the documentation:


StanleyGoldman:
We have actual progress methods now...

PR #552, opened 2017-12-29
Async install of Portable Git

StanleyGoldman:
## Presently

The installation of portable git is currently a synchronous process initiated from `ApplicationManagerBase`.

https://github.com/github-for-unity/Unity/blob/a1ec1cbce2066f76d772cf35484e7d14f9876ddb/src/GitHub.Api/Application/ApplicationManagerBase.cs#L46-L82

https://github.com/github-for-unity/Unity/blob/a1ec1cbce2066f76d772cf35484e7d14f9876ddb/src/GitHub.Api/Application/ApplicationManagerBase.cs#L139-L158

It's responsibilities are as follows.
This process is currently linear & synchronous, running on the main thread.

- Setup Git
  - Determine where git is installed
    - Get the value from settings, if it exists use that
    - If the previous step did not produce a path
      - Check for the existence of our Portable Git folder
        - Check that the git exec is present
        - Check that the git lfs exec is present
        - Check that the git-lfs exec matches the expected MD5
        - Attempt to extract git and git-lfs if the zip is present
        - If this process is successful use that path
    - If the previous step did not produce a path
      - And we are not on windows and `/usr/local/bin/git` exists, use that
    - If the previous step did not produce a path
      - run the system `where`/`which` command to find git.
  - Set `Environment.GitExecutablePath`
  - If on Windows, check that the credential helper is set to `wincred`
- Initialize the `User` with `GitClient`
- Restart Repository
- Initialize UI

## Pull Request

https://github.com/github-for-unity/Unity/blob/e6762b5147c9365bec04b5a4754b0a3d25287584/src/GitHub.Api/Application/ApplicationManagerBase.cs#L46-L117

In order to make this process asynchronous.
Several async tasks are created, upon their completion they choose which tasks to continue with.

### Async flow

- `DEFINE` After git is setup
  - Restart Repository
  - Initialize UI
- `DEFINE` Windows Credential setup
  - Set windows credential helper
  - `START` **"After git is setup"**
- `DEFINE` After path is determined
  - If windows `START` **"Windows Credential setup"**
  - Otherwise `START` **"After git is setup"**
- `DEFINE` Locate system git
  - If successful `START` **"After path is determined"**
  - Else throw `ERROR`
- `DEFINE` Locate or install portable git
  - If successful `START` **"After path is determined"**
  - Else `START` **"Locate system git"**

`START` **"Locate or install portable git"**

### Git Installer
This process is now async and performs the following tasks:
- Check for the existence of our Portable Git folder
  - Check that the extracted directory exists
  - Use a file list MD5 to verify that all expected files are present
    - Using a file list MD5 because this would be run on every startup and a full MD5 takes a few seconds
    - It would also be overkill because a full MD5 is performed on the zip files output
  - Perform a full MD5 on Git LFS
- If Portable Git is not discovered to be intact
  - Check if it can use portable git based on OS (we don't ship portable git for mac currently, but for when we do)
  - Uses `UnzipTask` to unzip both zip files and perform a full MDS on the results
  - Creates a combined output and moves it to the target directory
  - Cleanup after itself
  
### UI Changes
  - Changes to `UserSettingsView` and `InitProjectView` to remain disabled until git is ready


StanleyGoldman:
I realized I have one regression, which if you read my documentation you will catch.
I'm missing a check against settings to see if the person has chosen a git.

StanleyGoldman:
Fixed

PR #559, opened 2018-01-04
Specifying some string CultureInfo settings

StanleyGoldman:
**Note:** This pull request targets `fixes/code-analysis` #458
   
Working through different code analysis issues in small pull requests.
In this pull request I'm specifying `CultureInfo` where possible.


StanleyGoldman:
"Current Culture" vs "Invariant Culture"...

Current Culture uses the culture from where the application is accessed. Often used for display. i.e. think of the accents on french letters.

Invariant Culture uses none of that. Used for calculation.

PR #561, opened 2018-01-04
Async git setup rollup

StanleyGoldman:
Composed of:
- [x] #545
> - Created `DownloadTask`
>   - A task that downloads a file with functionality to support resume
>   - If specified the task will also verify the MD5 of a file
>   - If specified the task will perform retries if there was an issue with the download
> - Created `DownloadTestTask`
>   - A task that downloads text from a url
- [x] #549
> - Creating a lazily loaded singleton of `ZipHelper`
> - Creating an `UnzipTask` and a basic integration test that shows it works
> - Removing delegate from `GitInstaller` and using the singleton `ZipHelper` instead. 
- [x] #550
> Adding functionality to MD5 a folder In order to validate the contents.
- [x] #563
> It's an often needed function and takes the burden of handling it repeatedly.
- [x] #552 
> ### Async flow
> 
> - `DEFINE` After git is setup
>   - Restart Repository
>   - Initialize UI
> - `DEFINE` Windows Credential setup
>   - Set windows credential helper
>   - `START` **"After git is setup"**
> - `DEFINE` After path is determined
>   - If windows `START` **"Windows Credential setup"**
>   - Otherwise `START` **"After git is setup"**
> - `DEFINE` Locate system git
>   - If successful `START` **"After path is determined"**
>   - Else throw `ERROR`
> - `DEFINE` Locate or install portable git
>   - If successful `START` **"After path is determined"**
>   - Else `START` **"Locate system git"**
> - `DEFINE` Check settings for git path
>   - If successful `START` **"After path is determined"**
>   - Else `START` **"Locate or install portable git"**
> 
> `START` **"Check settings for git path"**
> 
> ### Git Installer
> This process is now async and performs the following tasks:
> - Check for the existence of our Portable Git folder
>   - Check that the extracted directory exists
>   - Use a file list MD5 to verify that all expected files are present
>     - Using a file list MD5 because this would be run on every startup and a full MD5 takes a few seconds
>     - It would also be overkill because a full MD5 is performed on the zip files output
>   - Perform a full MD5 on Git LFS
> - If Portable Git is not discovered to be intact
>   - Check if it can use portable git based on OS (we don't ship portable git for mac currently, but for when we do)
>   - Uses `UnzipTask` to unzip both zip files and perform a full MDS on the results
>   - Creates a combined output and moves it to the target directory
>   - Cleanup after itself
>   
> ### UI Changes
>   - Changes to `UserSettingsView` and `InitProjectView` to remain disabled until git is ready
- [x] #562
> - Removing functionality from `GitInstaller` to locate zip files from local assemblies
> - Added functionality to `GitInstaller`
>   - Ability to specify local paths to zip files for install
>   - If local paths are not specified, `GitInstaller` pre-pends a `DownloadTask` to the `UnzipTask`
> - Added functionality to unit tests to local zip files from local assemblies, as we don't need tests downloading files we have locally.
- [x] #553
> - Adding the `LoadingView` to `Window` and using it until we detect that a git executable has been set
> - Centering text in the `LoadingView` waiting for the sprite


shana:
Pulling this one in to keep iterating on it and preparing a release.

PR #583, opened 2018-02-22
Loading animation when the window is busy

shana:
Switching the loading view into an overlay drawn over an existing view, with a loading spinner and a progress bar.

There's still some bugs here, needs testing.

spinner


shana:
This has been merged into shana/fix-initialization #604 and fixed there, shana/fix-refresh #608 and 1.0rc/master #610 include it. When that gets merged into master, this will be too.

PR #585, opened 2018-02-22
Download portable git zip files to a known location

StanleyGoldman:
- Moving downloaded zip files to the AppData folder 
- Preventing download of a file when it exists

Currently, the DownloadTask will download and append to a file, even if it already exists and
passes MD5.

In order to fix this:
- Check if the file exists
 - If there is no validation hash delete the file
 - If there is a validation hash, check if it matches, if so bypass the
 downloading an additional check

### DIFF
--- a/src/GitHub.Api/Installer/GitInstaller.cs
+++ b/src/GitHub.Api/Installer/GitInstaller.cs
@@ -24,10 +24,13 @@ namespace GitHub.Unity
 
         private readonly bool onWindows;
 
-        public GitInstallDetails(NPath targetInstallPath, bool onWindows)
+        public GitInstallDetails(NPath applicationDataPath, bool onWindows)
         {
             this.onWindows = onWindows;
-            var gitInstallPath = targetInstallPath.Combine(ApplicationInfo.ApplicationName, PackageNameWithVersion);
+
+            PluginDataPath = applicationDataPath.Combine(ApplicationInfo.ApplicationName);
+
+            var gitInstallPath = PluginDataPath.Combine(PackageNameWithVersion);
             GitInstallationPath = gitInstallPath;
 
             if (onWindows)
@@ -55,6 +58,7 @@ namespace GitHub.Unity
                 : gitInstallRoot.Combine("libexec", "git-core", GitLfsExecutable);
         }
 
+        public NPath PluginDataPath { get; }
         public NPath GitInstallationPath { get; }
         public string GitExecutable { get; }
         public NPath GitExecutablePath { get; }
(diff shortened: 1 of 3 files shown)


src/tests/IntegrationTests/Installer/GitInstallerTests.cs
```diff
@@ -35,6 +35,7 @@ public override void TestFixtureTearDown()
         }
 
         [Test]
+        [Category("DoNotRunOnAppVeyor")]
```
StanleyGoldman: I'm not sure why all of a sudden this test seems to be stopped up in AppVeyor

src/GitHub.Api/Installer/GitInstaller.cs
```diff
@@ -25,10 +25,13 @@ class GitInstallDetails
 
         private readonly bool onWindows;
 
-        public GitInstallDetails(NPath targetInstallPath, bool onWindows)
+        public GitInstallDetails(NPath applicationDataPath, bool onWindows)
         {
             this.onWindows = onWindows;
-            var gitInstallPath = targetInstallPath.Combine(ApplicationInfo.ApplicationName, PackageNameWithVersion);
+
+            PluginDataPath = applicationDataPath.Combine(ApplicationInfo.ApplicationName);
```
shana: I think we already have this in the environment, UserCachePath?

src/tests/IntegrationTests/Installer/GitInstallerTests.cs
StanleyGoldman: I isolated this change in #595 and whatever is wonky can be fixed there.

src/GitHub.Api/Installer/GitInstaller.cs
StanleyGoldman: 👍

src/GitHub.Api/Tasks/DownloadTask.cs
```diff
@@ -99,30 +99,53 @@ protected virtual string RunDownload(bool success)
...
 
-                    using (var destinationStream = fileSystem.OpenWrite(Destination, FileMode.Append))
+                    var fileExistsAndValid = false;
+                    if (Destination.FileExists())
                     {
-                        result = Utils.Download(Logger, Url, destinationStream,
-                            (value, total) =>
+                        if (ValidationHash == null)
+                        {
+                            Destination.Delete();
```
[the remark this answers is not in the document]
StanleyGoldman: The following testing scenario is what lead me to this decision.
1. Clear downloaded zip files and delete portable git.
2. Start GitHub for Unity and let it download zip and install portable git.
3. Delete portable Git.
4. Start GitHub for Unity
  a. It downloads the same content and double the size of the zip.
  b. Zip fails MD5 check
  c. Zip download again correctly.
  d. Portable Git is installed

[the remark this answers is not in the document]
StanleyGoldman: So yea, I made the determination that a file should only be resume-able if you are able to validate it's end result.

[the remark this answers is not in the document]
shana: > a. It downloads the same content and double the size of the zip.
I think this may be making things less efficient. How about this sequence:

### There's no files downloaded

1. Download md5
2. Download file with temporary name (rename operations are cheap as long as the file doesn't change locations). Something like realname.extension.partial
3. When done, rename file
4. Run md5 check
5. If md5 check fails, delete file and start download process from scratch

### There's a .partial file

1. Resume downloading of .partial file.
2. When done, rename file
3. Run md5 check
4. If md5 check fails, delete file and start download process from scratch

### There's a completed downloaded file

1. Run md5 check
2. If md5 check fails, delete file and start download process from scratch

> a file should only be resume-able if you are able to validate it's end result.
The testing scenario you describe includes an md5 check so I'm not sure how that scenario relates to your conclusion here. In any case, if we name a partial download clearly, we know it's a partial and we can request a resume regardless of whether there's an md5 to back it up.

If we don't have an md5 to verify a file, then we need to trust that if a file looks complete, it's probably complete, otherwise every time Unity restarts we'll be downloading things over and over again, which is really wasteful. We should avoid downloading things without an md5 to verify integrity in general anyways.

PR #590, opened 2018-02-23
GitInstaller should prefer local resources before downloading

StanleyGoldman:
- Added functionality to `GitInstaller` to check locally for git and git lfs zip files
- If one or both are found the `GitInstaller` will download the respective files
- Afterwards it resumes as expected.

### DIFF
--- a/src/GitHub.Api/Application/ApplicationManagerBase.cs
+++ b/src/GitHub.Api/Application/ApplicationManagerBase.cs
@@ -72,9 +72,22 @@ namespace GitHub.Unity
                         }
                     });
 
+                var archiveParentPath = NPath.CreateTempDirectory("git_zip_paths");
+                var gitArchiveFilePath = AssemblyResources.ToFile(ResourceType.Platform, "git.zip", archiveParentPath, Environment);
+                if (!gitArchiveFilePath.FileExists())
+                {
+                    gitArchiveFilePath = null;
+                }
+
+                var gitLfsArchivePath = AssemblyResources.ToFile(ResourceType.Platform, "git-lfs.zip", archiveParentPath, Environment);
+                if (!gitLfsArchivePath.FileExists())
+                {
+                    gitLfsArchivePath = null;
+                }
+
                 var applicationDataPath = Environment.GetSpecialFolder(System.Environment.SpecialFolder.LocalApplicationData).ToNPath();
                 var installDetails = new GitInstallDetails(applicationDataPath, true);
-                var gitInstaller = new GitInstaller(Environment, CancellationToken, installDetails);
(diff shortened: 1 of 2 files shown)


src/GitHub.Api/Installer/GitInstaller.cs
```diff
@@ -129,9 +130,48 @@ public void SetupGitIfNeeded(ActionTask<NPath> onSuccess, ITask onFailure)
...
             {
-                downloadFilesTask = CreateDownloadTask();
+                archiveParentPath = NPath.CreateTempDirectory("git_zip_paths");
+
+                if (gitArchiveFilePath == null)
+                {
+                    gitArchiveFilePath = AssemblyResources.ToFile(ResourceType.Platform, "git.zip", archiveParentPath, environment);
+                    if (!gitArchiveFilePath.FileExists())
+                    {
+                        gitArchiveFilePath = null;
```
StanleyGoldman: Should this file exists check be moved into `AssemblyResources.ToFile(...)`?

StanleyGoldman: I also was not sure if the usage of `AssemblyResources.ToFile(...)` be used here inside `GitInstaller` or the result sent into the constructor.

StanleyGoldman: > I also was not sure if the usage of AssemblyResources.ToFile(...) be used here inside GitInstaller or the result sent into the constructor.

Now that I say that out loud I feel the like call should be outside `GitInstaller`...

shana: Yeah, I agree it probably should be outside, this should only be concerned with extracting a zip given a file path.

src/GitHub.Api/Application/ApplicationManagerBase.cs
```diff
@@ -72,9 +72,22 @@ public void Run(bool firstRun)
...
 
+                var archiveParentPath = NPath.CreateTempDirectory("git_zip_paths");
+                var gitArchiveFilePath = AssemblyResources.ToFile(ResourceType.Platform, "git.zip", archiveParentPath, Environment);
+                if (!gitArchiveFilePath.FileExists())
+                {
+                    gitArchiveFilePath = null;
+                }
+
+                var gitLfsArchivePath = AssemblyResources.ToFile(ResourceType.Platform, "git-lfs.zip", archiveParentPath, Environment);
+                if (!gitLfsArchivePath.FileExists())
```
StanleyGoldman: This feels a bit better.. Maybe a static method in `GitInstaller` would be better than this...

StanleyGoldman: Ehh...

shana:
This has been manually ported to #609

PR #593, opened 2018-02-26
Unit tests that show Tree control maintains checked/collapsed state

StanleyGoldman:
- Adding tests to prove that the load function
  - Loads with collapsed nodes as expected
  - Loads with selected nodes as expected


StanleyGoldman:
So after my first foray into this, I'm able to reproduce the bug, I see that the load functionality works as expected. I suspect there is a serialization issue, but everything is the same as checked state.

Lobsterovich:
5+

Lobsterovich:
5+

shana:
There's a bug somewhere, but it's really hard to track down. I got it a couple of times, but it's not consistently reproducible.

collapsed-tree

StanleyGoldman:
Cool, i changed the labels on the bug.

StanleyGoldman:
@shana I think after #698 this bug is resolved. See if you can reproduce it.

StanleyGoldman:
i'm pretty sure it happens every other domain reload

StanleyGoldman:
but data that is created after the first domain reload, is persisted after the second domain reload

StanleyGoldman:
touching the file system twice and focusing unity once, which is one domain reload, as i understand it, also reproduces the issue

PR #595, opened 2018-02-27
GitInstall test seems erratic

StanleyGoldman:
- Ignoring test from AppVeyor, until we can figure this out


shana:
I'm rewriting all this stuff so we shouldn't need this

PR #596, opened 2018-02-27
Adding functionality to prevent changes to locked files

StanleyGoldman:
We want to be able to prevent users from modifying files that are locked. [UnityEditor.AssetModificationProcessor](https://docs.unity3d.com/ScriptReference/AssetModificationProcessor.IsOpenForEdit.html) gives us that functionality.

**Todo**:
- [x] Create a `LfsLocksModificationProcessor` that can match file locks to 
- [x] Determine the current git lfs user's username or identifier.
- [x] Fix the definition `GitLock.Default`


StanleyGoldman:
This is looking good so far, but I don't know how to identify the current LFS user. 

@ttaylor: Is there a `git lfs` command I can user to query the current lfs user.

ttaylorr:
> @ttaylorr: Is there a `git lfs` command I can user to query the current lfs user.

Hey, I think @shana and I figured this out this morning:

> I'm assuming the handle returned from the github api for the user will match the lfs user, is that accurate?

Which is correct.

shana:
This is ready, needs testing.

src/UnityExtension/Assets/Editor/GitHub.Unity/UI/LfsLocksModificationProcessor.cs
```diff
@@ -0,0 +1,107 @@
...
+                return true;
+            }
+        }
+
+        private static void RepositoryOnLocksChanged(CacheUpdateEvent cacheUpdateEvent)
+        {
+            if (!lastLocksChangedEvent.Equals(cacheUpdateEvent))
+            {
+                lastLocksChangedEvent = cacheUpdateEvent;
+                locks = repository.CurrentLocks;
```
StanleyGoldman: From what I've seen and read about `AssetModificationProcessor` the methods inside get called very often.
Should we create a dictionary from this list object to mitigate that?

shana: Yeah, probably a good idea to preformat the data here to make the lookups as fast as possible in the callbacks.

PR #599, opened 2018-02-27
Octorun Js

StanleyGoldman:
Due to the TLS changes recently introduced in github.com
We are providing a new means to access the GitHub API for the purposes of:
1. Account Authentication & Validation
2. Publishing
We are also providing the means to report usage metrics.

The means is via a node application written to support node 0.12.7.
This is the version of nodejs that ships with Unity 5.4.

We are also using a version of octokit/rest.js that has been transpiled to run on nodejs 0.12.
https://github.com/gr2m/octokit-rest-for-node-v0.12 provided to us by @gr2m (:wave: _**thanks**_)

There is `OctorunTask` which encapsulates environment variables required by octorun-js.
We still need to package octorun-js and deploy it in the target environment.


octorun/src/authentication.js
```diff
@@ -0,0 +1,84 @@
...
+            }
+            else {
+                onFailure(err)
+            }
+        }
+        else {
+            onSuccess(res.data.token);
+        }
+    });
+}
```
shana: `handleBasicAuthentication` and `handleTwoFactorAuthentication` look exactly the same except for an extra header in the request. Could this be deduped into one method with a flag?

octorun/src/octokit.js
```diff
@@ -0,0 +1,19 @@
+var Octokit = require('octokit-rest-for-node-v0.12');
+
+var createOctokit = function () {
+    return Octokit({
+        timeout: 0,
+        requestMedia: 'application/vnd.github.v3+json',
+        headers: {
+            'user-agent': 'octokit/rest.js v1.2.3'
```
shana: This needs to be the same string that we pass in C# as ProductHeaderValue - the name and version of the app. It should probably be set in the environment by C#, probably easiest that way.

src/GitHub.Api/Application/ApiClient.cs
```diff
@@ -198,22 +210,55 @@ public ApiClient(UriString hostUrl, IKeychain keychain, IGitHubClient githubClie
                 await ValidateKeychain();
                 await ValidateCurrentUserInternal();
 
-                GitHubRepository repository;
+                var uriString = keychain.Connections.First().Host;
+                var keychainAdapter = await keychain.Load(uriString);
+
+                var command = new StringBuilder("publish -r ");
+                command.Append(newRepository.Name);
```
shana: This should be surrounded with quotes. Not that there will be spaces in this name, but there might be other characters that could be interpreted by the shell and cause problems.

src/GitHub.Api/Application/ApiClient.cs
```diff
@@ -198,22 +210,55 @@ public ApiClient(UriString hostUrl, IKeychain keychain, IGitHubClient githubClie
...
+
+                if (!string.IsNullOrEmpty(newRepository.Description))
+                {
+                    command.Append(" -d ");
+                    command.Append(newRepository.Description);
```
shana: This should be in quotes.

src/GitHub.Api/Application/ApiClient.cs
```diff
@@ -198,22 +210,55 @@ public ApiClient(UriString hostUrl, IKeychain keychain, IGitHubClient githubClie
...
+                }
+
                 if (!string.IsNullOrEmpty(organization))
                 {
-                    logger.Trace("Creating repository for organization");
+                    command.Append(" -o ");
+                    command.Append(organization);
```
shana: This should be in quotes.

src/GitHub.Api/Application/ApiClient.cs
```diff
@@ -198,22 +210,55 @@ public ApiClient(UriString hostUrl, IKeychain keychain, IGitHubClient githubClie
...
+                var ret = await octorunTask.StartAwait();
+
+                if (ret.Count == 0)
                 {
-                    logger.Trace("Creating repository for user");
+                    throw new ApiClientException("Publish failed");
+                }
 
-                    repository = (await githubClient.Repository.Create(newRepository)).ToGitHubRepository();
+                if (ret[0] == "success")
```
shana: Since everyone that's running `OctorunTask` is going to be doing exactly the same `if (ret[0] == "success")` to check for status codes and output, it's probably a good idea if we have an actual object with the data returned from OctorunTask instead of a list of strings perhaps? Manually checking strings is very error prone. Maybe just having a static helper method that everyone calls to convert the output of OctorunTask to something that's properly typed would be the the easiest way, no need to change OctorunTask at all.

src/GitHub.Api/Platform/DefaultEnvironment.cs
```diff
@@ -131,6 +132,23 @@ public NPath GitExecutablePath
...
+
+        public NPath NodeJsExecutablePath
+        {
+            get
+            {
+                if (!nodeJsExecutablePath.IsInitialized)
+                {
+                    nodeJsExecutablePath = IsWindows
+                        ? UnityApplication.Parent.Combine("Data", "Tools", "nodejs", "node.exe")
+                        : UnityApplication.Combine("Contents", "Tools", "nodejs", "node");
```
shana: This should probably be using the value returned by `UnityEditor.EditorApplication. applicationContentsPath` (https://docs.unity3d.com/ScriptReference/EditorApplication-applicationContentsPath.html). We should probably add a `UnityApplicationContents` property to the environment so we can use it here. I'm fine with doing that on a separate PR, this isn't a blocker

src/GitHub.Api/Tasks/BaseOutputProcessor.cs
```diff
@@ -126,4 +126,28 @@ protected override bool ProcessLine(string line, out NPath result)
             return true;
         }
     }
+
+    class LastResultOutputProcessor<T> : BaseOutputProcessor<T>
```
shana: This isn't being used, is it?

octorun/src/octokit.js
StanleyGoldman: So I'm looking at what value is being used currently...

https://github.com/github-for-unity/Unity/blob/b3850c6d6206eef04fbbeeb3cb2d1d1fa940a0ab/src/GitHub.Api/Application/ApplicationConfiguration.cs#L14

https://github.com/github-for-unity/Unity/blob/b3850c6d6206eef04fbbeeb3cb2d1d1fa940a0ab/src/GitHub.Api/Application/ApplicationInfo.cs#L6-L14

We are using `ApplicationInfo.ApplicationSafeName`, should it not be? `ApplicationInfo.ApplicationName`

src/GitHub.Api/Application/ApiClient.cs
StanleyGoldman: 👍

src/GitHub.Api/Application/ApiClient.cs
StanleyGoldman: 👍

src/GitHub.Api/Application/ApiClient.cs
StanleyGoldman: 👍

src/GitHub.Api/Application/ApiClient.cs
StanleyGoldman: Addressed in #616

src/GitHub.Api/Tasks/BaseOutputProcessor.cs
StanleyGoldman: Nope, I'll remove it.

octorun/src/octokit.js
shana: Yeah, it should, dunno why safe name is there like that.

shana: Actually it should always be `GitHubUnity` so you can just set `ApplicationSafeName` to that

src/GitHub.Api/Platform/DefaultEnvironment.cs
StanleyGoldman: Tracked in #617

octorun/src/authentication.js
```diff
@@ -0,0 +1,84 @@
...
+        headers: {
+            "X-GitHub-OTP": twoFactor
+        }
+    }, function (err, res) {
+        if (err) {
+            if (lockedRegex.test(err.message)) {
+                onFailure("Account locked.")
+            }
+            else if (badCredentialsRegex.test(err.message)) {
+                onFailure("Bad credentials.")
```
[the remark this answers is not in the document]
StanleyGoldman: Tracked in #618

octorun/src/authentication.js
StanleyGoldman: Tracked in #619

octorun/src/octokit.js
StanleyGoldman: Tracked in #620

PR #604, opened 2018-03-06
Task/Cache Jenga 2: Fixing initialization and git installation and other fixes

shana:
Built on top of shana/fix-tasks #609
Followed by #608 

- Merging in the loading spinner from #583 
- deduping and killing unused code
- fixing the git installation sequence
- replaying cache invalidation requests that happen before RepositoryManager is set (because data might be requested and invalidated before we've had a chance to finish the initialization)
- fixing the process task begin/end sequence (it could have thrown before OnEnd is called)


shana:
Closing this because the fixes to get this green need to go on top of shana/fix-refresh #608

PR #608, opened 2018-03-07
[do not merge] Task/Cache Jenga 3: Fixing cache invalidation and refresh

shana:
Built on top of shana/fix-tasks #609 

It includes, from shana/fix-initialization #604:

- Merging in the loading spinner from #583 
- deduping and killing unused code
- fixing the git installation sequence
- replaying cache invalidation requests that happen before RepositoryManager is set (because data might be requested and invalidated before we've had a chance to finish the initialization)
- fixing the process task begin/end sequence (it could have thrown before OnEnd is called)

And then it includes:

The cache needs to be refreshed whenever something happens on the filesystem, but there is a strong possibility that Unity will reload the domain before we have a chance to get the data back into the cache after issuing a git status or other commands that take a while to complete, since filesystem changes can often cause a recompilation.

In this PR, filesystem watcher events trigger cache invalidations that will then trigger git commands. If the new data is never received because Unity reloaded the domain, the next cache access will cause it to be refreshed, since it's been invalidated already (by resetting the UpdatedAt date)

This also changes the system so that nothing runs until the cache is accessed, making everything on demand., which means the assumption that events happen as soon as the repository managers starts up  is no longer valid.


src/GitHub.Api/Git/Repository.cs
```diff
@@ -38,24 +43,54 @@ public Repository(NPath localPath, ICacheContainer container)
...
+                    CurrentRemoteChanged?.Invoke(cacheUpdateEvent);
+                    CurrentBranchAndRemoteChanged?.Invoke(cacheUpdateEvent);
+                }},
+            };
+
             cacheContainer = container;
-            cacheContainer.CacheInvalidated += CacheContainer_OnCacheInvalidated;
-            cacheContainer.CacheUpdated += CacheContainer_OnCacheUpdated;
+            cacheContainer.CacheInvalidated += CacheHasBeenInvalidated;
+            cacheContainer.CacheUpdated += (cacheType, offset) => cacheUpdateEvents[cacheType](new CacheUpdateEvent(cacheType, offset));
```
StanleyGoldman: Nice.

src/GitHub.Api/Git/Repository.cs
```diff
@@ -309,273 +207,97 @@ private void CacheContainer_OnCacheInvalidated(CacheType cacheType)
...
         }
 
-        private GitBranch GetLocalGitBranch(ConfigBranch x)
+        private static GitBranch GetLocalGitBranch(string currentBranchName, ConfigBranch x)
         {
             var branchName = x.Name;
             var trackingName = x.IsTracking ? x.Remote.Value.Name + "/" + branchName : "[None]";
-            var isActive = branchName == CurrentBranchName;
-
+            var isActive = branchName == currentBranchName;
```
shana: BTW I did this because calling `CurrentBranchName` hits the cache and can trigger an invalidation request, which is not something we want to do when we're updating another cache.

src/GitHub.Api/Managers/Downloader.cs
```diff
@@ -67,7 +67,7 @@ class PairDownloader
             private readonly IFileSystem fs;
             private readonly CancellationToken cancellationToken;
 
-            private volatile int finishedTaskCount;
+            private int finishedTaskCount;
```
StanleyGoldman: I didn't notice this in the first pull request and I'm noticing it now. What makes these variables `volatile`? And why is that variable no longer?

src/GitHub.Api/Application/ApplicationManagerBase.cs
```diff
@@ -171,34 +191,34 @@ protected void SetupMetrics(string unityVersion, bool firstRun)
...
             if (Environment.IsWindows)
             {
-                var task = GitClient
-                    .GetConfig("credential.helper", GitConfigSource.Global)
-                    .Then((b, credentialHelper) => {
-                        if (string.IsNullOrEmpty(credentialHelper))
+                var credHelperTask = GitClient.GetConfig("credential.helper", GitConfigSource.Global);
+                credHelperTask.OnEnd += (thisTask, credentialHelper, success, exception) =>
+                    {
+                        if (!success || string.IsNullOrEmpty(credentialHelper))
```
StanleyGoldman: This pattern is much easier to understand.

src/GitHub.Api/Cache/CachingClasses.cs
```diff
@@ -0,0 +1,18 @@
...
+using GitHub.Unity;
+using System;
+using System.Collections.Generic;
+using System.Globalization;
+using System.Linq;
+using System.Runtime.Serialization;
+
+namespace GitHub.Unity
+{
+    sealed class RepositoryInfoCacheData : IRepositoryInfoCacheData
```
StanleyGoldman: Are you making a single unit out of these values? Is this a pattern we should be repeating in the cache classes?

src/GitHub.Api/Git/Repository.cs
```diff
@@ -309,273 +207,97 @@ private void CacheContainer_OnCacheInvalidated(CacheType cacheType)
...
-            new ActionTask(TaskManager.Instance.Token, () => {
-                CurrentChanges = gitStatus.Entries;
-                CurrentAhead = gitStatus.Ahead;
-                CurrentBehind = gitStatus.Behind;
-            }) { Affinity = TaskAffinity.UI }.Start();
+            taskManager.RunInUI(() =>
+            {
+                cacheContainer.GitStatusEntriesCache.Entries = gitStatus.Entries;
+                cacheContainer.GitTrackingStatusCache.Ahead = gitStatus.Ahead;
+                cacheContainer.GitTrackingStatusCache.Behind = gitStatus.Behind;
```
StanleyGoldman: Yes, I see how we would want a single structure to handle the updates here and below.

src/GitHub.Api/Managers/Downloader.cs
shana: `volatile` is a hint to the compiler that this field will be touched by multiple threads, so it won't reorder operations for optimization purposes. Normally the compiler could try to optimize things by reordering operations that aren't related to each other in a method:

```
a = true;
if (b == 2) {}
```

These two operations can be run in any order and there's no guarantee that the compiler will do them in this particular order, since it doesn't matter what the order is, they aren't dependent on each other. But when you have fields that are touched from multiple places, two operations that are seemingly unrelated might suddenly have side effects that affect each other in ways that the compiler can't predict, and reordering them might introduce subtle bugs. Like for instance some other thread changing the value of `b` depending on whether `a` is true or not - then the order of setting `a` and reading `b` would be important. This isn't a real world example but it's the gist of it.

The reason I added `volatile` is because `finishedTaskCount` gets written to every time a task finishes. The reason I took it out is because  `Interlocked.Increment` already guarantees atomicity of the operation in a threaded environment and that's the only time this field is touched, so marking it as `volatile` is pointless (and the compiler helpfully points this out).

src/GitHub.Api/Cache/CachingClasses.cs
shana: There's two reasons I'm doing this. The first reason is the major one:

When updating fields in the cache, every individual field set will save it to disk and it will raise the cache updated event. Now each cache object has multiple fields, and they're all updated at the same time when the data comes in from git, which means every set will trigger a save and a cache update event. All these events coming from a single cache object will contain partially updated information until the very last field update. The views will get these events and render them multiple times. That's not ideal.

Given that a cache object has multiple pieces of data but represents one atomic event unit, then each update to a cache object must be made atomically with all the information we have for it in one go, and then saved as one piece and one update event raised from it when it's done.

To avoid raising multiple events per cache update and still have the cache save itself and set its updatedDate fields properly, we can't update its fields one by one, we have to update it all in one go (if we set fields individually, then the cache has no way of knowing when the update is done, which means the caller is now responsible for manually telling the cache that the update is complete, which is an extra burden that would be nice not to have).

That means that we have to gather all the information and pass it in a single blob to the cache object, for it to update itself in one go and raise a single event. For that to work in a nice way, we have to have a way of representing the blob of data that's inside the cache. This is what I'm doing here as an experiment: splitting the functionality and the data of the RepositoryInfoCache, so that we can have objects that represent only the data inside it, and we don't have to have separate objects to represent this data (or some other untyped data structure to represent blobs of data).


The other reason is that in order to test things properly I have to have properly working cache objects on the testing side. Since they're all scriptable objects, I can't reference them directly in the standalone tests, which is really really annoying.  I could create some Unity tests, but we can't run those in CI (they run in Unity itself).

So I ended up copy pasting all of the cache objects into IntegrationTests and removing all the Unity specific stuff from them, which is a paaaaaaain 'cause now we have to keep them in sync. We can't use the trick of subclassing them from GitHub.Api because they already subclass ScriptableObject. Sooooo, I'm trying to find ways to pull out as much as possible from the cache objects into GitHub.Api and only leave the bare minimum in the cache objects in GitHub.Unity. Maybe all they do is redirect all their calls into another object that implements all the logic and lives somewhere we can use in tests? Not sure yet, but this is what this is experimenting with.

shana: If this works, yeah, I'll change all the other cache classes to use this pattern.

src/GitHub.Api/Application/ApplicationManagerBase.cs
```diff
@@ -127,17 +146,18 @@ public void RestartRepository()
...
             if (Environment.RepositoryPath != null)
             {
-                repositoryManager = Unity.RepositoryManager.CreateInstance(Platform, TaskManager, GitClient, ProcessManager, Environment.FileSystem, Environment.RepositoryPath);
+                repositoryManager = Unity.RepositoryManager.CreateInstance(Platform, TaskManager, GitClient, Environment.FileSystem, Environment.RepositoryPath);
                 repositoryManager.Initialize();
-                Environment.Repository.Initialize(repositoryManager);
+                Environment.Repository.Initialize(repositoryManager, TaskManager);
                 repositoryManager.Start();
-                Logger.Trace($"Got a repository? {Environment.Repository}");
+                Environment.Repository.Start();
```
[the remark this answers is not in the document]
shana: It's super effective!

PR #611, opened 2018-03-07
Removing Octokit

StanleyGoldman:
This targets #599 and does the work of removing Octokit from the application.
This includes libraries and a special solution used for debugging Octokit.
The majority of code changes to the application are removing `IGitHubClient` and credentials objects.


src/GitHub.Api/Application/ApiClient.cs
```diff
@@ -214,12 +211,12 @@ public ApiClient(UriString hostUrl, IKeychain keychain, IGitHubClient githubClie
...
                 var command = new StringBuilder("publish -r ");
-                command.Append(newRepository.Name);
+                command.Append(repositoryName);
 
-                if (!string.IsNullOrEmpty(newRepository.Description))
+                if (!string.IsNullOrEmpty(description))
                 {
                     command.Append(" -d ");
-                    command.Append(newRepository.Description);
+                    command.Append(description);
```
shana: This needs quotes around it for the command line to parse properly

src/GitHub.Api/Application/ApplicationConfiguration.cs
```diff
@@ -11,19 +10,13 @@ static ApplicationConfiguration()
         {
             var executingAssembly = typeof(ApplicationConfiguration).Assembly;
             AssemblyName = executingAssembly.GetName();
-            ProductHeader = new ProductHeaderValue(ApplicationInfo.ApplicationSafeName, AssemblyName.Version.ToString());
```
shana: Not related directly to this PR, but you should make sure this information is sent along with the requests in the node app, so you might need to set it in the environment as well.

src/GitHub.Api/Authentication/LoginManager.cs
```diff
@@ -83,108 +73,59 @@ class LoginManager : ILoginManager
...
-                {
-                    auth = new ApplicationAuthorization(password);
-                }
-                else
-                {
-                    await keychain.Clear(host, false);
-                    return new LoginResultData(LoginResultCodes.Failed, Localization.LoginFailed, host);
-                }
+                await keychain.Clear(host, false);
+                return new LoginResultData(LoginResultCodes.Failed, Localization.LoginFailed, host);
```
shana: This is missing the other error conditions like locked out and the workaround for enterprise, isn't it?

src/GitHub.Api/Application/ApiClient.cs
StanleyGoldman: Addressed in #599

src/GitHub.Api/Application/ApplicationConfiguration.cs
StanleyGoldman: Tracked in #617

src/GitHub.Api/Authentication/LoginManager.cs
StanleyGoldman: Tracked in #617

src/GitHub.Api/Application/ApplicationConfiguration.cs
shana: This should be fixed in #599

PR #613, opened 2018-03-08
Dealing with a few more NPath changes

StanleyGoldman:


src/GitHub.Api/Application/ApplicationManagerBase.cs
```diff
@@ -60,7 +60,7 @@ public void Run(bool firstRun)
                 var initEnvironmentTask = new ActionTask<NPath>(CancellationToken, (_, path) => InitializeEnvironment(path)) { Affinity = TaskAffinity.UI };
                 var findExecTask = new FindExecTask("git", CancellationToken)
                     .FinallyInUI((b, ex, path) => {
-                        if (b && path != null)
+                        if (b && path != NPath.Default)
```
shana: This can be `path.IsInitialized` instead of comparing with `NPath.Default`, which is way faster

Issue #621, opened 2018-03-09
Octorun JS Missing Enterprise Login Workaround


StanleyGoldman:
Breakout of #617

- [x] Implement missing enterprise workaround
https://github.com/github-for-unity/Unity/pull/611#discussion_r173294989

As explained here, we need a specific status code to indicate that an enterprise user should utilize their password as a token.

https://github.com/github-for-unity/Unity/blob/0aa51f7d14cb57841c07865133a834b26453da8b/src/GitHub.Api/Authentication/LoginManager.cs#L254-L265

Issue #619, opened 2018-03-09
Octorun JS refactor authentication methods


StanleyGoldman:
Breakout of #617

- [x] `handleBasicAuthentication` and `handleTwoFactorAuthentication` can be refactored
https://github.com/github-for-unity/Unity/pull/599#pullrequestreview-102650159

In Octorun the authentication methods in `authentication.js` are very similar and can be consolidated.

https://github.com/github-for-unity/Unity/blob/d05f08b78f9b209f9f7969400aeb2892d5eb2da8/octorun/src/authentication.js#L11
https://github.com/github-for-unity/Unity/blob/d05f08b78f9b209f9f7969400aeb2892d5eb2da8/octorun/src/authentication.js#L46

Issue #618, opened 2018-03-09
Octorun JS replace IsCustom and use specific values


StanleyGoldman:
Breakout of #617

- [x] Replace `IsCustom` with specific values for each "status code" that can be returned
  - [x] Only output status codes or raw error messages
https://github.com/github-for-unity/Unity/pull/599#discussion_r173454720
  - [x] Add status code for 2FA code failure
https://github.com/github-for-unity/Unity/pull/611#discussion_r173294989

In the `OctorunResult` object we allow for a user to parse the status when it is not a known "Success/Error" status. We can do better by being deterministic about the values returned by the javascript and parsed for in c#.

https://github.com/github-for-unity/Unity/blob/7860c089ea46cfa0a20387c11b3079ba35adba83/src/GitHub.Api/Tasks/OctorunTask.cs#L107-L127

That means we should only output status when the octokit return values are parseable by Octorun. If not Octorun should return the raw error to the C# code.

StanleyGoldman:
Octokit returns the same 2FA error message if the code is wrong

PR #624, opened 2018-03-09
Refactoring Octorun output and authentication.js

StanleyGoldman:
Fixes: #618
Fixes: #619
Fixes: #621

- Adding specific error statuses and checking for them specifically in C#
- Refactoring authentication.js to reduce repetition
- Removing unneeded exceptions used for message passing
- Implementing the missing enterprise workaround
- Consolidated TryLogin and TryContinueLogin


src/GitHub.Api/Authentication/LoginManager.cs
```diff
@@ -174,25 +174,33 @@ string password
...
             {
-                throw new Exception(string.Join(Environment.NewLine, ret.Output));
+                return new LoginResultData(LoginResultCodes.Failed, "Failed.", host, ret.Output[0]);
             }
 
-            throw new Exception("Authentication failed");
+            return new LoginResultData(LoginResultCodes.Failed, "Failed.", host);
         }
 
-        private async Task<string> TryContinueLogin(
```
shana: This method and the `TryLogin` method are exactly the same except for an error message, a flag and the extra line written to the process. Maybe dedupe this too?

StanleyGoldman: Consolidated

PR #640, opened 2018-03-24
Refactoring methods to prevent double loading of data

StanleyGoldman:

Depends on:
- [ ] #652 

The validation methods `ValidateKeychain` and `ValidateCurrentUserInternal` load the requisite data and validate them. The methods that are calling these methods perform the same tasks to load the same data items and continue.

In order to reduce the re-querying the validation methods now load the data, validate the data and returns the data if valid or throw an exception if invalid as previously expected.

Refactoring method `LoadKeychainInternal`/`ValidateKeychain` to `GetValidatedKeychainAdaper`
Refactoring method `ValidateCurrentUserInternal`/`GetCurrentUserInternal` to `GetValidatedGitHubUser`

### DIFF
--- a/src/GitHub.Api/Application/ApiClient.cs
+++ b/src/GitHub.Api/Application/ApiClient.cs
@@ -86,20 +86,15 @@ namespace GitHub.Unity
             Guard.ArgumentNotNull(onSuccess, nameof(onSuccess));
             try
             {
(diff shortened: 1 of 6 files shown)


StanleyGoldman:
I'm going to put a bit more into the explanation and resolution of this problem.

Issue #650, opened 2018-03-27
The plugin fails to initialize LFS


valyard:
The plugin silently fails when trying to create a new repo or change settings.
Log contains this error: `Exception on concurrent thread: 1 lfs version
GitHub.Unity.ProcessException: git: 'lfs' is not a git command. See 'git --help'.`

I am starting Unity 2017.3.1 with an empty project and pressing "initialize repo" in the plugin window. Nothing seems to happen, though the .git folder in the project is created.

First of all, it would be nice to have some kind of error reported to the UI when a failure happens.

I'm not sure about my local git and lfs install, but it works from command line and other tools.
```
valentin-mac:git-test valyard$ which git
/usr/local/bin/git
valentin-mac:git-test valyard$ git lfs version
git-lfs/2.3.3 (GitHub; darwin amd64; go 1.9.1)
```

Where is the plugin running `git` from?

[github-unity.log](https://github.com/github-for-unity/Unity/files/1849930/github-unity.log)

StanleyGoldman:
Hey @valyard 
Thanks for testing out our stuff..
I'm still reading your log file, but it seems to be using git found at `/usr/bin/git`.

StanleyGoldman:
Hey @valyard 
Do you have a file at `~/.config/GitHubUnity/settings.json`?
What is in it?

valyard:
`~/.config/GitHubUnity/` is empty.
`/usr/bin/git` might be a link to `/usr/local/bin/git` since `/usr/bin/git lfs version` also returns `git-lfs/2.3.3 (GitHub; darwin amd64; go 1.9.1)`.

StanleyGoldman:
Hey @valyard so we are still debugging, this bug is reproduced on a few of our machines here. So it should not take us long to resolve.

StanleyGoldman:
**TLDR**: I've isolated the problem to how we computer the `PATH` environment variable in `ProcessEnvironment`

I'm able to reproduce this on my mac. I have some updated information.
I was interested in capturing what the `PATH` was being set to. I added some logging to...

https://github.com/github-for-unity/Unity/blob/0c3b816f3fdeaeae8648a79707e34e9e0f8c041b/src/GitHub.Api/Platform/ProcessEnvironment.cs#L60-L73

I got this...
```
/usr/bin:/usr::/usr:/usr/bin:/bin:/usr/sbin:/sbin:
```

[...]

PR #658, opened 2018-03-28
ValidateGitInstall may fail in error

### LINKED ISSUE #656: Settings - Save path on macOS does nothing
When there's no portable git on the system and it's just detecting whatever git is in the system, changing the git path in the settings view makes the view disabled but doesn't actually save anything.

StanleyGoldman:

Fixes #656

Capturing a failure from ValidateGitInstall.

On Windows after picking something completely irrelevant.

image

On my Mac currently suffering from #650 

image


StanleyGoldman:
Yea, that works in theory, after some mac testing it doesn't work out when LFS is not found.

PR #659, opened 2018-03-29
Fixing path variable calculation

StanleyGoldman:
### Problems
1. After some investigation I realized that the PATH variables were being calculated/changed when calling an external git. Calculating a PATH variable for our internal portable git makes sense, but if the user is presenting us with a git setup, we should assume that the Git setup is functional on the system.
2. Compared to the default PATH variable a Mac user has when he opens the terminal, the PATH variable we get through Unity/Mono appears incomplete. That is because OSX uses an additional script `/usr/libexec/path_helper` to augment the PATH variable. [src](https://stackoverflow.com/questions/41315582/incomplete-path-variable).
3. When you open a shell in GitHub for Unity, `path_helper` is executed. As a result the path variables when we open a terminal window does not match the path variables when we are calling Git. Which explains why users were able to run `git lfs ...` in the terminal we opened for them when we couldn't do it in code.

### Fixes
1. In `ApplicationManagerBase.Run()` we are calling `path_helper` if we are on a mac, and updating `Environment.Path` with the result.
2. Being more careful about how we craft our PATH variable.
3. If we are going to make decisions based on if we are using portable git or not, we need to let the users "Reset" back to the portable git we provide. #660 
4. We also cannot use our calculated path when attempting to find the system git, as that will cloud our results. #661
5. Being sure the path we craft for console, actually matches what we are using when running git commands. #662

Fixes: #650 

Depends on: 
- [ ] #660
- [ ] #661
- [ ] #662


StanleyGoldman:
**TLDR**: I've isolated the problem to how we computer the `PATH` environment variable in `ProcessEnvironment`

I'm able to reproduce this on my mac. I have some updated information.
I was interested in capturing what the `PATH` was being set to. I added some logging to...

https://github.com/github-for-unity/Unity/blob/0c3b816f3fdeaeae8648a79707e34e9e0f8c041b/src/GitHub.Api/Platform/ProcessEnvironment.cs#L60-L73

I got this...
```
/usr/bin:/usr::/usr:/usr/bin:/bin:/usr/sbin:/sbin:
```

Meanwhile on my normal shell.. _(Minus some Ruby/homebrew stuff...)_
```
/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin:
```

In code I forcibly set that value in place and I was able to initialize the repo.

StanleyGoldman:
I did a bit more digging with some log messages.

https://github.com/github-for-unity/Unity/blob/e49517fb63d4bb97a962e6cdf49f68daeb2a166d/src/GitHub.Api/Platform/ProcessEnvironment.cs#L31-L37

`ProcessEnvironment.Configure()` attempts to craft a `PATH` environment around this installation.
But it is using the current Git installation as it's frame of reference, not the git installation being presented to it.

StanleyGoldman:
On my mac I have git & git-lfs installed via homebrew, which creates a symlink `/usr/local/bin/git` which points to `/usr/local/Cellar/git/[version]/git`.

```bash
C02L939KF6T6:~ stanleygoldman$ ls -la /usr/local/bin/git
lrwxr-xr-x  1 stanleygoldman  admin  28 Mar 29 10:12 /usr/local/bin/git -> ../Cellar/git/2.16.3/bin/git
```
If a user attempts to select that path using the file browser, the file browser seems to resolve the symlink and give us the real path.

A savvier user might attempts to to save a git path by typing in `/usr/local/bin/git`, in that scenario we attempt to validate that git path, we do not resolve the symlink.

StanleyGoldman:
So thinking about this further. If we are providing Git to the user, it makes sense that we need to supplement the environment's `PATH`, but if the user is supplying their own git... I believe we should assume they have it installed correctly, and the environment's `PATH` should work correctly.

StanleyGoldman:
This `dontSetupGit` variable could be useful..

https://github.com/github-for-unity/Unity/blob/e49517fb63d4bb97a962e6cdf49f68daeb2a166d/src/GitHub.Api/Platform/ProcessEnvironment.cs#L25-L26

StanleyGoldman:
There is also no way to go back to the "default" setting.
The lack of a git path in a user's settings leads to the usage of the internal portable git path.
After a user changes this setting for the first time, there is no way to go back to the way things were.
A user could attempt to choose the portable git path, but that will just save this value to the user's settings.

If my comment above is correct, whereby we only modify the `PATH` if the user has changed it. This is a scenario that would fail. In this scenario we should modify the `PATH` because the user is attempting to use our portable git.

StanleyGoldman:
Also, note what happens here with opening a command window on a mac...

https://github.com/github-for-unity/Unity/blob/e49517fb63d4bb97a962e6cdf49f68daeb2a166d/src/GitHub.Api/OutputProcessors/ProcessManager.cs#L85-L92

https://github.com/github-for-unity/Unity/blob/e49517fb63d4bb97a962e6cdf49f68daeb2a166d/src/GitHub.Api/Platform/ProcessEnvironment.cs#L75-L79

When running a command through `ProcessEnvironment` we set the path to the value we calculate.
When opening a command shell with `ProcessManager` we take what was given to us by `ProcessEnvironment` and we **_pre-pend_** it to the users `PATH`.

Which explains this scenario...

image

We cannot use LFS when we run commands, but the shell we open is able to.

Here is that shell's `PATH`
```bash
bash-3.2$ echo $PATH
/usr/bin:/usr::/usr:/usr/bin:/bin:/usr/sbin:/sbin::/Users/stanleygoldman/.rvm/gems/ruby-2.1.5-github/bin:/Users/stanleygoldman/.rvm/gems/ruby-2.1.5-github@global/bin:/Users/stanleygoldman/.rvm/rubies/ruby-2.1.5-github/bin:/Users/stanleygoldman/.rbenv/shims:/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin:/Library/Frameworks/Mono.framework/Versions/Current/Commands:/Users/stanleygoldman/.rvm/bin:/Users/stanleygoldman/.rvm/bin
```

StanleyGoldman:
I believe most of the battle is to be fought here.

https://github.com/github-for-unity/Unity/blob/e49517fb63d4bb97a962e6cdf49f68daeb2a166d/src/GitHub.Api/OutputProcessors/ProcessManager.cs#L32-L48

Here we choose what executable to run, if none is chosen we choose git.
We also have the choke point on `dontSetupGit` here as well.

StanleyGoldman:
Also, the "Find Install" function is not working correctly... It does not specify `dontSetupGit`. As a result on Windows the find task returns the currently configured git installation. I can only wonder what it does on a mac.

https://github.com/github-for-unity/Unity/blob/977b32eb756e3f37b83bbbf66cba52b505e964b2/src/UnityExtension/Assets/Editor/GitHub.Unity/UI/GitPathView.cs#L122-L123

As compared to what happens during mac installation.

https://github.com/github-for-unity/Unity/blob/977b32eb756e3f37b83bbbf66cba52b505e964b2/src/GitHub.Api/Installer/GitInstaller.cs#L44-L47

StanleyGoldman:
This seems to give an explanation about why things are different...

https://stackoverflow.com/questions/41315582/incomplete-path-variable

Digging through `/etc/profile` I find the call to `/usr/libexec/path_helper`.
Calling it on my machine I get this...

```bash
C02L939KF6T6:GitHub stanleygoldman$ /usr/libexec/path_helper
PATH="/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin:/Library/Frameworks/Mono.framework/Versions/Current/Commands:/Users/stanleygoldman/.rvm/gems/ruby-2.1.5-github/bin:/Users/stanleygoldman/.rvm/gems/ruby-2.1.5-github@global/bin:/Users/stanleygoldman/.rvm/rubies/ruby-2.1.5-github/bin:/Users/stanleygoldman/.rbenv/shims:/Users/stanleygoldman/Projects/go/bin:/Users/stanleygoldman/.rvm/bin"; export PATH;
```

Which coincidentally matches the `PATH` that I get in the terminal in the comments above... 🤔

src/GitHub.Api/Application/ApplicationManagerBase.cs
```diff
@@ -54,14 +53,34 @@ public void Run(bool firstRun)
         {
             Logger.Trace("Run - CurrentDirectory {0}", NPath.CurrentDirectory);
 
+            ITask<string> getMacEnvironmentPathTask;
+            if (Environment.IsMac)
+            {
+                getMacEnvironmentPathTask = new SimpleProcessTask(CancellationToken, "bash".ToNPath(), "-c \"/usr/libexec/path_helper\"")
+                           .Configure(ProcessManager)
```
shana: This should probably set `dontUseGit: true` since we just want to run bash

PR #661, opened 2018-03-29
Specifying dontUseGit when finding git

StanleyGoldman:
Fixes: #664

Part of the problems identified in #659 

When attempting to use the system's `where`/`which` command we should probably not modify the environment `PATH` variable. Especially when attempting to look for the system git.


src/UnityExtension/Assets/Editor/GitHub.Unity/UI/GitPathView.cs
```diff
@@ -120,7 +120,7 @@ public override void OnGUI()
                         CheckEnteredGitPath();
 
                         new FindExecTask("git", Manager.CancellationToken)
-                            .Configure(Manager.ProcessManager)
+                            .Configure(Manager.ProcessManager, false, true)
```
shana: This should just be `.Configure(Manager.ProcessManager, dontSetupGit: true)` for readability and future-proofing.

PR #678, opened 2018-04-05
Fixing octorun output processing

StanleyGoldman:
The result of the usage API call is not a string but an object that has toString.
We should just output that


Lobsterovich:
Cool

PR #682, opened 2018-04-09
Fix an infinite cache invalidation loop

shana:
Repro:

- Pull or Push or otherwise trigger a filesystem event that causes the plugin to update the current remote or current branch information when that information hasn't changed since the last time it was updated.

This causes an infinite cache invalidation loop that will hang Unity.

### DIFF
--- a/src/UnityExtension/Assets/Editor/GitHub.Unity/ApplicationCache.cs
+++ b/src/UnityExtension/Assets/Editor/GitHub.Unity/ApplicationCache.cs
@@ -136,6 +136,7 @@ namespace GitHub.Unity
         [SerializeField] private string initializedAtString = DateTimeOffset.MinValue.ToString(Constants.Iso8601Format);
         [NonSerialized] private DateTimeOffset? lastUpdatedAtValue;
         [NonSerialized] private DateTimeOffset? initializedAtValue;
+        [NonSerialized] private bool isInvalidating;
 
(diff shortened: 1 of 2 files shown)


src/UnityExtension/Assets/Editor/GitHub.Unity/ApplicationCache.cs
```diff
@@ -161,21 +162,27 @@ public bool ValidateData()
...
 
         protected void SaveData(DateTimeOffset now, bool isChanged)
         {
             var isInitialized = IsInitialized;
             isChanged = isChanged || !isInitialized;
 
-            InitializedAt = !isInitialized ? now : InitializedAt;
-            LastUpdatedAt = isChanged ? now : LastUpdatedAt;
+            InitializedAt = !isInitialized || InitializedAt == DateTimeOffset.MinValue ? now : InitializedAt;
+            LastUpdatedAt = isChanged || LastUpdatedAt == DateTimeOffset.MinValue ? now : LastUpdatedAt;
```
shana: This is the fix for the infinite invalidation loop - if the data hasn't changed, `LastUpdated` is never set back to a valid value, causing the `ValidateData` to think the cache has timed out and causing an invalidation, which comes back here and doesn't set the `LastUpdatedAt` field, etc ad nauseaum.

PR #713, opened 2018-05-04
Per session usage metrics of GitHub for Unity

StanleyGoldman:
- Added an InstanceId to uniquely identify runs of GitHub for Unity
- Using that InstanceId to group usage metrics

### DIFF
--- a/src/GitHub.Api/Metrics/UsageTracker.cs
+++ b/src/GitHub.Api/Metrics/UsageTracker.cs
@@ -17,19 +17,21 @@ namespace GitHub.Unity
         private readonly NPath storePath;
         private readonly ISettings userSettings;
         private readonly IMetricsService metricsService;
-        private readonly string guid;
+        private readonly string userId;
         private readonly string unityVersion;
+        private readonly string instanceId;
         private Timer timer;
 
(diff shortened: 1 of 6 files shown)


src/GitHub.Api/Metrics/UsageTracker.cs
```diff
@@ -154,10 +156,7 @@ private void RunTimer(int seconds)
 
         private Usage GetCurrentUsage(UsageStore usageStore)
         {
-            var usage = usageStore.Model.GetCurrentUsage(ApplicationConfiguration.AssemblyName.Version.ToString(), unityVersion);
-            usage.Dimensions.Lang = CultureInfo.InstalledUICulture.IetfLanguageTag;
-            usage.Dimensions.CurrentLang = CultureInfo.CurrentCulture.IetfLanguageTag;
```
shana: Why remove these two lines? Is this being done somewhere else?

StanleyGoldman: Hmm... Seems I lost some code...

shana: Ah, cool!

PR #717, opened 2018-05-04
git fetch should also fetch tags

### LINKED ISSUE #572: git fetch should also fetch tags
`git fetch` doesn't fetch tags that aren't directly reachable by current branch heads. Long-lived release or maintenance branches are not normally merged back to master and tags created on them won't be fetched normally. We should do ~`git fetch --all`~`git fetch --tags` to ensure all tags are fetched.

EDIT: we want the --tags flag, not the --all flag

ThomasAunvik:
Fixes #572 

### Description of the Change

Adds `--all` in the arguments whenever you want to fetch all in `GitFetchTask`, third parameter, 

### Benefits

Fetches Tags

### Possible Drawbacks

Possible longer time to finish fetch task.

### DIFF
--- a/src/GitHub.Api/Git/GitClient.cs
+++ b/src/GitHub.Api/Git/GitClient.cs
@@ -276,7 +276,7 @@ namespace GitHub.Unity
         {
             //Logger.Trace("Fetch");
 
-            return new GitFetchTask(remote, cancellationToken, true, processor)
+            return new GitFetchTask(remote, cancellationToken, true, true, processor)
                 .Configure(processManager);
         }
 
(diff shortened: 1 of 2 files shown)


shana:
@ThomasAunvik Looking back at the issue and more at git, it turns out the solution in the original issue is incorrect. `--all` is for fetching all the remotes, I should have written `--tags` for fetching all the tags (even non-reachable ones). Sorry, my bad!

Also, the new flag should come after the prune flag in the method signature.

shana:
@ThomasAunvik Also thank you for tackling this stuff, it's really helpful! 🎉

PR #720, opened 2018-05-07
Feature: Update check

### LINKED ISSUE #702: Add functionality to notify users of updates
Analysis of our metrics shows that our users are not updating GitHub for Unity, but we are getting new users on newer versions of GitHub for Unity.

We would like to implement some version check that would help notify users to newer versions of the plugin.

shana:
Fixes #702 

This PR adds a UI to inform users that there's a new update for them. In the process, it introduces a package json with the following example format:

```
{"md5":"ea5d5a38a6b9e9bc2b10011602c65a0d", "url":"localhost:55555/unity/releases/github-for-unity-99.2.0-beta1.unitypackage","releaseNotes":null,"releaseNotesUrl":null,"message":null,"executableMd5":null, "version":"99.2.0-beta1"}
```

This json format is also now used for serving the `git.zip` and `git-lfs.zip` files. The various release notes and message fields are used to show information in the UI. The executableMd5 field is used to validate existing installations of git and git-lfs.

It also introduces a `TheVersion` class that represents a version, with string parsing and comparison capabilities. Versions are in the format Major.Minor.Patch.Build, and Major is the only mandatory component. Any part of Minor, Patch or Build can have alphanumeric characters, eg: 1.0-beta, which will compare/sort lower than version 1.0 (but higher than 0.x)

The extension checks for updates the first time Unity runs. It allows the user to skip an update.

Update checks to git.zip and git-lfs.zip are done once a day.

There are a couple of new scripts to generate package files and for generating the `latest.json` update feed.

This also stops bundling git and git-lfs with the extension.

Dark skin:
image

Light skin:
image


src/GitHub.Api/Application/ApplicationManagerBase.cs
```diff
@@ -54,79 +46,34 @@ public void Run(bool firstRun)
             Logger.Trace("Run - CurrentDirectory {0}", NPath.CurrentDirectory);
             isBusy = true;
 
-            var endTask = new ActionTask<GitInstaller.GitInstallationState>(CancellationToken,
-                    (_, state) => InitializeEnvironment(state))
-                { Affinity = TaskAffinity.UI };
-
-            ITask<string> setExistingEnvironmentPath;
-            if (Environment.IsMac)
+            var thread = new Thread(obj =>
```
StanleyGoldman: I'm surprised to see `new Thread` here. What was the reason?

StanleyGoldman: Ah, you were tired of the knot of task chains..

src/GitHub.Api/Git/GitConfig.cs
```diff
@@ -306,35 +306,38 @@ public Section(string name, string description = null)
...
             public string GetString(string key)
             {
-                return this[key];
+                return this[key].First();
             }
 
             public int GetInt(string key)
             {
-                var value = this[key];
+                var value = TryGetString(key);
```
StanleyGoldman: This method should probably be renamed to `TryGetInt`.

src/GitHub.Api/Git/GitConfig.cs
```diff
@@ -306,35 +306,38 @@ public Section(string name, string description = null)
...
                 int result = 0;
                 int.TryParse(value, out result);
                 return result;
             }
 
             public float GetFloat(string key)
             {
-                var value = this[key];
+                var value = TryGetString(key);
```
StanleyGoldman: This method should probably be renamed to `TryGetFloat`.

src/GitHub.Api/Helpers/AssemblyResources.cs
```diff
@@ -33,7 +33,22 @@ public static NPath ToFile(ResourceType resourceType, string resource, NPath des
             if (stream != null)
                 return destinationPath.Combine(resource).WriteAllBytes(stream.ToByteArray());
 
-            return environment.ExtensionInstallPath.Combine(type, os, resource).Copy(destinationPath.Combine(resource));
+            // check the GitHub.Api assembly
```
StanleyGoldman: I've looked at this method before and I've always wondered about some of the paths it takes and returns from.
Can you tell me about this function so I can add comments to it?

src/GitHub.Api/Installer/GitInstaller.cs
```diff
@@ -11,206 +12,183 @@ class GitInstaller
...
-            else
-            {
-                startTask = new FuncTask<GitInstallationState>(cancellationToken, () =>
-                    {
-                        return VerifyPortableGitInstallation();
-                    })
-                { Name = "Git Installation - Verify" };
+                    Logger.Trace("Using git install path from settings: {0}", gitExecutablePath);
+                    state.GitExecutablePath = gitExecutablePath.Value;
+                    state.GitIsValid = true;
```
StanleyGoldman: If someone provides us with a custom git. Shouldn't we search for Git LFS locally on the system here. Currently it seems that if someone gives us a copy of git we are still going to use our own Git LFS.

src/GitHub.Api/Installer/GitInstaller.cs
```diff
@@ -11,206 +12,183 @@ class GitInstaller
...
-                    return;
-                }
+                return VerifyMacGit(state);
+            }
 
-                var downloadZipTask = DownloadZipsIfNeeded(installState);
-                downloadZipTask.OnEnd += ExtractPortableGit;
-                thisTask.Then(downloadZipTask);
-            };
+            state = VerifyPortableGitInstallation(state);
```
StanleyGoldman: If the git path has chosen from settings, this method will overwrite the choice made by settings.
We should definitely attempt to find GitLFS and if both are valid, return `state` before this line of code.

src/GitHub.Api/Primitives/TheVersion.cs
```diff
@@ -0,0 +1,267 @@
+using System;
+using System.Text.RegularExpressions;
+
+namespace GitHub.Unity
+{
+    public struct TheVersion : IComparable<TheVersion>
```
StanleyGoldman: Naming things is hard...

src/GitHub.Api/Installer/GitInstaller.cs
shana: Oh yes, I knew I was missing something

src/UnityExtension/Assets/Editor/GitHub.Unity/UI/GitPathView.cs
```diff
@@ -239,10 +239,13 @@ private void ValidateAndSetGitInstallPath(string value)
...
 
-                var gitInstaller = new GitInstaller(Environment, EntryPoint.ApplicationManager.ProcessManager,
-                    EntryPoint.ApplicationManager.TaskManager);
+                var gitInstaller = new GitInstaller(Environment, Manager.ProcessManager,
+                    TaskManager, Manager.SystemSettings);
 
-                gitInstaller.SetupGitIfNeeded()
+                new FuncTask<GitInstaller.GitInstallationState>(TaskManager.Token, () =>
+                    {
+                        return gitInstaller.SetupGitIfNeeded();
```
StanleyGoldman: Nice. Basically no change from this end.

src/UnityExtension/Assets/Editor/GitHub.Unity/UI/Window.cs
```diff
@@ -60,14 +60,20 @@ public static void GitHub_CommandLine()
...
+        public static void GitHub_SelectWindow()
+        {
+            var window = Resources.FindObjectsOfTypeAll(typeof(Window)).FirstOrDefault() as Window;
+            Selection.activeObject = window;
+        }
+
+        [MenuItem("GitHub/Restart")]
+        public static void GitHub_Restart()
+        {
+            EntryPoint.Restart();
```
StanleyGoldman: Restart without a domain restart?

src/UnityExtension/Assets/Editor/GitHub.Unity/GitHub.Unity.asmdef
```diff
@@ -0,0 +1,8 @@
+{
+    "name": "GitHub.Unity",
+    "references": [],
+    "includePlatforms": [
+        "Editor"
+    ],
+    "excludePlatforms": []
```
StanleyGoldman: We talked about this quickly in person. Can you remind my brain what this is for?

src/UnityExtension/Assets/Editor/GitHub.Unity/UpdateCheck.cs
```diff
@@ -0,0 +1,230 @@
...
+            {
+                if (success)
+                {
+                    try
+                    {
+                        Package package = result.ReadAllText().FromJson<Package>(lowerCase: true, onlyPublic: false);
+                        TheVersion current = TheVersion.Parse(ApplicationInfo.Version);
+                        TheVersion newVersion = package.Version;
+
+                        TheVersion[] versionsToSkip = EntryPoint.ApplicationManager.UserSettings.Get<TheVersion[]>(Constants.SkipVersionKey);
```
StanleyGoldman: Why did you choose to store an array and not the last skipped version?

shana: Good point, the latest skipped version is enough

src/UnityExtension/Assets/Editor/GitHub.Unity/GitHub.Unity.asmdef
shana: Unity will compile all files in a directory with a asmdef in their own DLL (UnityTests are then compiled in a separate DLL from this one). It's not really very relevant, we have our own csproj files for building the DLLs, but it's nice to test.

src/UnityExtension/Assets/Editor/GitHub.Unity/UI/Window.cs
shana: Yes. It's going to leak but it's only for our use anyway.

PR #725, opened 2018-05-09
Dedicated view to Locked files

StanleyGoldman:
Depends on 
- [x] #730


StanleyGoldman:
I found two bugs and updated the gif.

donokuda:
**Idea for design polish.** Trying to keep UI changes to a minimum, but here's what I was thinking:

image

- I combined the lock owner and timestamp to one line and modified it to read like a sentence. Having the owner's name by itself felt too ambiguous.
- Added the lock icon on top of the asset icon. The goal is so that people will be able to understand what the icon means when they see it in the project view.
- (Minor tweak) Scale the icon so that it fits in a 32x32 container, per sizing recommendations in the  [Unity HIG](https://unitytech.github.io/unityeditor-hig/topics/checklist.html)
 
I also made an attempt to spec out the spacing and font sizes for each list item:
image

-----

Bonus earlier exploration, same idea but with smaller icons: 

image

I was also going to try out a table-like view, but it seemed like too big of a departure of what we got and there wouldn't be enough horizontal space for all information.

PR #731, opened 2018-05-11
Listing locks in json format so they include the lock time

shana:
The --json flag provides more data regarding the locks that are taken, including the time that the lock was taken. This should also fix some parsing problems some users were getting and make everything simpler.


src/GitHub.Api/Helpers/SimpleJson.cs
```diff
@@ -1495,7 +1499,9 @@ protected virtual object SerializeEnum(Enum p)
         protected virtual bool TrySerializeKnownTypes(object input, out object output)
         {
             bool returnValue = true;
-            if (input is DateTime)
+            if (input is NPath)
+                output = input.ToString();
```
StanleyGoldman: That's good to know

PR #738, opened 2018-05-11
Validate git installations by version and fix setting up custom gits

shana:
Since we know which git and git lfs versions we want and we already check them in the settings view, switch to using version checks instead of md5 everywhere. If we find a good enough version, use it.

Make sure we do update checks once a day for git and git lfs.

Go and find lfs if we have a git custom path on all OSes

Make sure we restart the repository manager and everything else that relies on git when we switch gits, since outputs may change.


src/GitHub.Api/Application/ApplicationManagerBase.cs
```diff
@@ -51,47 +51,70 @@ public void Run()
...
+                                state.GitLfsIsValid = version >= Constants.MinimumGitLfsVersion;
+                            }
+                        }
+                    }
 
-                if (runInstallers)
-                {
-                    Environment.OctorunScriptPath = new OctorunInstaller(Environment, TaskManager)
-                        .SetupOctorunIfNeeded();
+                    if (!skipInstallers)
```
StanleyGoldman: I think you mean to set the value of `skipInstallers` with the result in `state`.

src/GitHub.Api/Application/ApplicationManagerBase.cs
```diff
@@ -170,28 +193,43 @@ private void SetupGit(GitInstaller.GitInstallationState state)
...
+
+                    GitClient.Init().RunWithReturn(true);
+                    GitClient.LfsInstall().RunWithReturn(true);
+                    AssemblyResources.ToFile(ResourceType.Generic, ".gitignore", targetPath, Environment);
+                    AssemblyResources.ToFile(ResourceType.Generic, ".gitattributes", targetPath, Environment);
+                    assetsGitignore.CreateFile();
+                    GitClient.Add(filesForInitialCommit).RunWithReturn(true);
+                    GitClient.Commit("Initial commit", null).RunWithReturn(true);
+                    Environment.InitializeRepository();
+                    UsageTracker.IncrementProjectsInitialized();
```
StanleyGoldman: Do you want to do this an an `ActionTask` exeuction, "fire and forget" style?

shana: The whole repo initialization takes a while, so doing this in thread is not a big deal I think.

PR #749, opened 2018-05-15
Removing WebTimeout Setting

StanleyGoldman:
This setting used to control the web timeout of Octokit.NET.
We could use it to control the timeout of  other things, but it's not currently connected.
Probably best to just remove it.


shana:
`ApplicationConfiguration.WebTimeout` is being used for web connections, you can see in https://github.com/github-for-unity/Unity/blob/a44bd0a7cc69941606b50ad080b2a10db9901a15/src/GitHub.Api/Managers/Downloader.cs#L145

PR #750, opened 2018-05-15
Updating ApplicationConfiguration webTimeout from settings

StanleyGoldman:
As a result of my hasty decisions in #749.

I also realized that the ApplicationConfiguration.WebTimeout was never initialized with the value from settings.

### DIFF
--- a/src/GitHub.Api/Application/ApplicationManagerBase.cs
+++ b/src/GitHub.Api/Application/ApplicationManagerBase.cs
@@ -42,13 +42,7 @@ namespace GitHub.Unity
             Platform = new Platform(Environment);
 
             LogHelper.TracingEnabled = UserSettings.Get(Constants.TraceLoggingKey, false);
-
-            int webTimeout;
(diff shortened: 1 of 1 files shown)


src/GitHub.Api/Application/ApplicationManagerBase.cs
```diff
@@ -42,6 +42,13 @@ protected void Initialize()
             Platform = new Platform(Environment);
 
             LogHelper.TracingEnabled = UserSettings.Get(Constants.TraceLoggingKey, false);
+
+            int webTimeout;
+            if (int.TryParse(UserSettings.Get(Constants.WebTimeoutKey), out webTimeout))
```
shana: settings will do the conversion for you - `UserSettings.Get<int>(Constants.WebTimeoutKey)`

StanleyGoldman: Fixed

PR #760, opened 2018-05-16
Moving the toolbar

StanleyGoldman:
Moving the toolbar from the `HistoryView` to `Window`

Depends on:
- [ ] #759


shana:
Did some tweaks so it now looks like this:

image

PR #781, opened 2018-05-21
.gitignore generation when non-existant.

### LINKED ISSUE #683: Create .gitignore on non-unity created repo
When you create a repo from for example Github Desktop and put the project folder inside that git, the Github for Unity wont ask you to add a .gitignore, since a repo already exists.

When you try to commit without the default Unity .gitignore you get this error on github desktop:
image
But when you try to commit in Github for Unity, nothing happens, summary and description gets removed.

If you could ask the player to add a .gitignore to the project folder if not found, else you wont be able to use git at all.

You can select a .gitignore file from the Github Desktop and that will add automaticly, but i made a mistake and forgot to add the .gitignore file to the project folder which made me get that error.

ThomasAunvik:
Fixes #683

### Description of the Change

This button generates .gitignore file in the project folder when pressed.

Suggesting a design change.


shana:
Yeah, this is definitely something we need, and I agree on the API change, I just need to think a bit on how this should be communicated in the UI. There's also the .gitattributes file, which is important for LFS (and arguably even more important than the .gitignore)

PR #788, opened 2018-05-23
Allow lock IDs to be strings, some servers return that instead of ints

### LINKED ISSUE #786: Parsing lock list fails on certain servers
Certain lfs server implementations use a hash for the lock id instead of a number, which causes our parsing to fail:

```
180522-11:31:49.089 ERROR [129] <LocksOutputProcessor>
Failed to parse lock line [{"id":"2f9cfde9c159d50e235cc1402c3e534b0bf2198afb20760697a5f9b07bf04fb3","path":"Assets/Scenes/Levels/Level1.unity","owner":{"name":"Owner"},"locked_at":"2018-05-22T18:31:53Z"}]
System.FormatException: Input string was not in a correct format.
```

shana:
Fixes #786


shana:
When merging this, please do a Squash and merge.

PR #803, opened 2018-05-29
Adding some ways of asking for a password, aka SSH support

### LINKED ISSUE #244: SSH support
We need to decide what level of SSH support we're going to have for the first major release:

1. Full - loading keys in ssh-agent, doing auth with it, etc, on all OSs (haha)
2. Partial - Falling to alternate https remotes if the current remote is ssh and doing all git operations over those alternate remotes
3. Minimal - Making sure we don't crash if the repo has an ssh remote. We trust that any git operations will work and won't actively engage in authentication or other key handling of any kind. Make sure all codepaths do not assume https

shana:
Fixes #244 

SSH_ASKPASS/GIT_ASKPASS will ask for a password if there isn't one. If the command
line asks for a password, use a custom UI to ask for it. I haven't tested the custom UI at all because haha I can't get it to trigger, processes that execute with redirected input don't ask for things on the command line. But for windows, at least, SSH_ASKPASS will ask for a password (though it won't cach it apparently).

This also moves reading process output to a thread because if the process asks for input without us noticing, it will hang forever and hang Unity in the process when it wants to kill the domain. This checks if we need to shutdown every second and aborts if needed.


ThomasAstrakhan:
I had to change my environment variable to make SSH askpass working on windows 10

ibytergj:
Hi what is the status of using ssh for GitHub for Unity?

PR #804, opened 2018-05-29
Adding the ability to see diffs of files and folders using the configured diff viewer

shana:
Unity has a preference to configure a diff viewer, so we can show diffs of the files that have been changed.

### DIFF
--- a/src/GitHub.Api/IO/FileSystem.cs
+++ b/src/GitHub.Api/IO/FileSystem.cs
@@ -46,12 +46,14 @@ namespace GitHub.Unity
         void WriteLines(string path, string[] contents);
 
         char DirectorySeparatorChar { get; }
+        string GetProcessDirectory();
     }
 
 
     public class FileSystem : IFileSystem
     {
         private string currentDirectory;
+        private string processDirectory;
 
         public FileSystem()
         { }
@@ -62,7 +64,7 @@ namespace GitHub.Unity
         /// <param name="directory">Current directory</param>
         public FileSystem(string directory)
         {
-            currentDirectory = directory;
+            processDirectory = currentDirectory = directory;
         }
 
         public void SetCurrentDirectory(string directory)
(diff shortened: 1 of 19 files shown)


src/GitHub.Api/IO/FileSystem.cs
```diff
@@ -242,6 +244,11 @@ public string GetCurrentDirectory()
             return Directory.GetCurrentDirectory();
         }
 
+        public string GetProcessDirectory()
+        {
+            return Directory.GetCurrentDirectory();
```
StanleyGoldman: Dont you mean to return `processDirectory` here?

src/GitHub.Api/Tasks/ActionTask.cs
```diff
@@ -221,16 +221,16 @@ class FuncTask<T, TResult> : TaskBase<T, TResult>
         protected Func<bool, T, TResult> Callback { get; }
         protected Func<bool, Exception, T, TResult> CallbackWithException { get; }
 
-        public FuncTask(CancellationToken token, Func<bool, T, TResult> action)
-            : base(token)
+        public FuncTask(CancellationToken token, Func<bool, T, TResult> action, Func<T> getPreviousResult = null)
```
StanleyGoldman: I couldn't find any scenarios where `getPreviousResult` is set to a value. Is this a bugfix?

StanleyGoldman: Mmm.. spoke too soon

src/GitHub.Api/IO/FileSystem.cs
shana: No, this is correct. The `processDirectory` field is not needed.

The process directory is whatever `Directory.GetCurrentDirectory()` returns, and our `GetCurrentDirectory()` "lies" about which directory is actually current. 

`FileSystem.GetCurrentDirectory` returns the repository root path, and `FileSystem.GetProcessDirectory` returns the Unity project path (which is what Unity sets the current directory to when it loads a project.

Most of the times these will be the same, but if the repository path is initialized somewhere else, we want NPath to always point to that as the current directory and operate based on the assumption that all paths start at the repository root (because that's where we always run our processes from).

beetbeet:
where and how to set diff viewer?

PR #807, opened 2018-05-29
ProjectWindow Initialization cleanups

shana:
Clean up null checks and initialization of the project window, so it doesn't throw if things race.


src/UnityExtension/Assets/Editor/GitHub.Unity/UI/ProjectWindowInterface.cs
```diff
@@ -11,18 +11,20 @@ class ProjectWindowInterface : AssetPostprocessor
...
-        private static readonly List<string> guidsLocks = new List<string>();
         private static IApplicationManager manager;
         private static bool isBusy = false;
         private static ILogging logger;
         private static ILogging Logger { get { return logger = logger ?? LogHelper.GetLogger<ProjectWindowInterface>(); } }
         private static CacheUpdateEvent lastRepositoryStatusChangedEvent;
         private static CacheUpdateEvent lastLocksChangedEvent;
-        private static IRepository Repository { get { return manager.Environment.Repository; } }
+        private static IRepository Repository { get { return manager != null ? manager.Environment.Repository : null; } }
+        private static bool IsInitialized { get { return Repository != null && Repository.CurrentRemote.HasValue; } }
```
StanleyGoldman: I understand why this is named `IsInitialized` because of how it is used in `EnsureInitialized`, but it could probably be better named as `HasRepositoryAndRemote`.

PR #828, opened 2018-06-12
Reducing the authentication scope requested by GitHub for Unity

StanleyGoldman:
While writing documentation I realized that we could reduce the token scope requested by GitHub for Unity.

Todo:
- [x] Create a new bundle of octorun


Monicasexy:
I agree it's a tad difficult!!

PR #881, opened 2018-08-01
Dealing with metafiles that show changes on build

StanleyGoldman:
Everytime I build these two files show up with just whitespace changes.

- Removing files from destination
- Updating `.gitignore`


shana:
This is weird, these files get copied from `lib/sfw/mac/libsfw.bundle.meta` and `lib/sfw/linux/libsfw.so.meta` during the build. The version that's committed in the Package directory is exactly the same as the one in the lib directory (we probably shouldn't have two copies of them, but oh well), why do you have modified files?

shana:
I'd rather update the .gitignore on the Package folder to ignore the files that are getting copied from lib (the `Assets/Plugins/GitHub/Editor/x64/` and `Assets/Plugins/GitHub/Editor/x86/` are already ignored, the others that are copied from lib should also have been ignored), these files should never change.

StanleyGoldman:
https://github.com/github-for-unity/Unity/commit/6cc72b76dc80d822f56c92d0afa9ae72fe025c0c

PR #889, opened 2018-08-17
Add support for the 4.6 profile side by side with 3.5 in 2018.x

### LINKED ISSUE #824: 2018.1 with .NET 4.5+ is broken
### Prerequisites

- Unity 2017+ 

### Description

Since Unity 2017 the .NET 4.X is available as experimental, on Unity 2018 it became fully supported: https://blogs.unity3d.com/2018/03/28/updated-scripting-runtime-in-unity-2018-1-what-does-the-future-hold/

However when you enable it inside Unity, it conflicts with the AsyncBridge related dependencies (TaskParallelLibrary, etc), showing errors that some classes already been defined in another assembly, since these classes are now present in the mscorlib.

### Steps to Reproduce

1. Enable Scripting Runtime Version .NET 4.X equivalent
2. Install GitHub for Unity 

### Additional Information

I've "fixed" this in a branch at my fork:
https://github.com/capyvara/Unity/tree/enhancements/dotnet-4.x

However I haven't submitted a PR since probably you're looking in supporting both .NET versions and I'm not sure how to better handle this.

shana:
Fixes #824

This PR adds support for the .net 4.6 profile by switching between the 3.5 and 4.5 versions of the DLLs and disabling importing of any dlls that are not needed for the target. This means that the System.Threading/AsyncBridge/ReadonlyCollections dlls we ship will only be loaded when running on the 3.5 profile.

GitHub.Api.45.csproj and GitHub.Unity.45.csproj are almost-exact copies of their original counterparts (just with the target framework and dll name changed, and irrelevant dependencies dropped).


shana:
A build of this branch is at https://github.com/github-for-unity/Unity/releases/tag/v1.0.3-beta1 for testing

src/UnityExtension/Assets/Editor/GitHub.Unity/ExtensionLoader/ExtensionLoader.cs
```diff
@@ -0,0 +1,86 @@
...
+                var filepath = path + file;
+                PluginImporter importer = AssetImporter.GetAtPath(filepath) as PluginImporter;
+                if (importer == null)
+                {
+                    Debug.LogFormat("GitHub for Unity: Could not find importer for {0}. Some functionality may fail.", filepath);
+                    continue;
+                }
+                if (importer.GetCompatibleWithEditor() != enable)
+                {
+                    importer.SetCompatibleWithEditor(enable);
```
StanleyGoldman: Wow.

Issue #899, opened 2018-08-27
After a certain number of changes, no changes are displayed locally


shafy:
### Description

After adding a couple of hundred files (importing assets), the changes just stop showing up completely (also the previously shown changed files are now gone from the "Changes" tab). 

This is logged if I add a new empty script:

```
180827-15:20:48.937 TRACE [ 1] <GitLocksCache>                     needsInvalidation isInitialized:True timedOut:True
180827-15:20:48.938 TRACE [181] <ProcessWrapper>                    Running '/usr/bin/git -c i18n.logoutputencoding=utf8 -c core.quotepath=false status -b -u --porcelain'
180827-15:20:48.948 TRACE [182] <ProcessWrapper>                    Running '/usr/local/bin/git-lfs locks --json'
```

### Steps to Reproduce

1. Open a new project
2. Install GitHub plugin
3. Publish repo
4. Import hundreds of asset files
5. "Changes" tab starts showing changed files
6. Import more files
7. "Changes" tab suddenly stops showing any changes at all (also older changed files are gone)

**Expected behavior:** Show all changed files.

**Actual behavior:** Shows 0 changed files.

**Reproduces how often:** 100%

StanleyGoldman:
Hey @shafy Thanks for testing our stuff out.

You don't happen to have a public repo I could clone to try to reproduce this error would you?
I tried a while back stress testing this thing with changes. I got pretty high without showing trouble...

Would you happen to have a closer count on how many? Also what types of files?
Maybe you can look at your commit history and let me know..

shafy:
Here, I made the repo public. It's non-sense test anyways :-) https://github.com/shafy/unity-kosmos-test-6

If you clone this project and add the GitHub plugin in Unity, adding more files doesn't show the changed files.

Regarding your questions:
- It was a whole folder, so I don't know which file type
- Ended up not committing and created a new repo through git command line

StanleyGoldman:
So I'm testing your project with Unity 2018.2 and GitHub for Unity 1.0.3.
I'm  able to make changes, but so far I'm unable to view the history of your project. I'm still looking into it.

shafy:
Cool. I used GitHub for Unity 1.0.2, Unity 2018.2.6f1 (right now I'm using this, not sure if I upgraded in the meantime to be honest) and macOS High Sierra 10.13.6

StanleyGoldman:
[...]

PR #921, opened 2018-10-03
Handling file names that look like they are a drive letter

StanleyGoldman:
In a user's repo the file `c:UsersOculusGoVideo.mp4` exists in history and trips up the log parsing process. The fix is being made in NPath (our IO library).

https://github.com/github-for-unity/Unity/blob/3a9593c3766e76dd3d6d7ec6d3b4aceac8441202/src/GitHub.Api/IO/NiceIO.cs#L114-L124

Tests have been added to verify this case works.

Fixes: #922
Related to #899 - https://github.com/github-for-unity/Unity/issues/899#issuecomment-426667124

### DIFF
--- a/src/tests/UnitTests/IO/GitObjectFactoryTests.cs
+++ b/src/tests/UnitTests/IO/GitObjectFactoryTests.cs
@@ -18,7 +18,7 @@ namespace UnitTests
                 CurrentDirectory = @"c:\Projects\UnityProject"
             });
 
-            var environment = NSubstitute.Substitute.For<IEnvironment>();
+            var environment = Substitute.For<IEnvironment>();
             environment.RepositoryPath.Returns(@"c:\Projects\UnityProject".ToNPath());
             environment.UnityProjectPath.Returns(@"c:\Projects\UnityProject".ToNPath());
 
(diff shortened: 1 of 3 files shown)


src/tests/UnitTests/IO/GitObjectFactoryTests.cs
```diff
@@ -0,0 +1,50 @@
...
+            });
+
+            var environment = Substitute.For<IEnvironment>();
+            environment.RepositoryPath.Returns(@"c:\Projects\UnityProject".ToNPath());
+            environment.UnityProjectPath.Returns(@"c:\Projects\UnityProject".ToNPath());
+
+            var gitObjectFactory = new GitObjectFactory(environment);
+            var gitStatusEntry = gitObjectFactory.CreateGitStatusEntry("c:UsersOculusGoVideo.mp4", GitFileStatus.Deleted);
+
+            Assert.AreEqual(@"c:\Projects\UnityProject\c:UsersOculusGoVideo.mp4", gitStatusEntry.FullPath);
```
queerviolet: WHO NAMES THEIR FILES THIS? 😂

StanleyGoldman: lol

PR #924, opened 2018-10-03
Fix bug to check parent file when meta is checked

### LINKED ISSUE #913: Committing new file by selecting its meta file does not commit the file itself
### Prerequisites
Tracing enabled, log file attached (but not much to see in it).
[github-unity.log](https://github.com/github-for-unity/Unity/files/2398910/github-unity.log)

### Description

When a new file is created, and the file is marked to be committed by selecting its meta file, the file itself will not be committed, even though the tree view shows it as also selected.

### Steps to Reproduce

1. Initialize GitHub for Unity, and a git repository
2. Create any new file in the project.
3. In the GitHub for Unity window click on the meta file of the newly created file. Notice that the newly created file is also marked as selected, as the meta file is a dependency of it.
4. Add commit message, and click on commit.

**Expected behavior:** All the files from the GitHub window that have their checkbox ticked in the tree view are (staged and) committed.

**Actual behavior:** Only the meta file is committed, without the file for which it was created.

**Reproduces how often:** 100% (Always)

### Additional Information
See attached "empty" Unity project, already in inconsistent state. Follow instructions to reproduce issue.
[GitCommitIssue.zip](https://github.com/github-for-unity/Unity/files/2398940/GitCommitIssue.zip)

Unity version is 2018.2.0f2, GitHub for Unity version is latest on Asset Store as of 2018-09-20.

It is probably a UI issue, inconsistency between what the UI logic shows and the list of files to be committed.

StanleyGoldman:
Fixes #913 

As a result of an item having it's check state changed in our tree lists, the check state of it's parents or children might need to be changed.

When performing this action from `TreeBase<TNode, TData>` we need to be sure to call `AddCheckedNode(...)` or `RemoveCheckedNode(...)` when those items are not folders. Those functions are overloaded and allow derived classes to set some additional state.

In order to do this cleanly, all nodes check states are now being changed through a central overloaded method.

https://github.com/github-for-unity/Unity/blob/d8d92c990278df1306754c22ed90de05193686ae/src/GitHub.Api/UI/TreeBase.cs#L303-L328

### DIFF
--- a/src/GitHub.Api/UI/TreeBase.cs
+++ b/src/GitHub.Api/UI/TreeBase.cs
@@ -332,7 +332,6 @@ namespace GitHub.Unity
             return results;
         }
 
-
         private void ToggleParentFoldersChecked(int idx, TNode node, bool isChecked)
         {
             while (true)
--- a/src/GitHub.Api/UI/TreeBase.cs
+++ b/src/GitHub.Api/UI/TreeBase.cs
@@ -261,7 +261,7 @@ namespace GitHub.Unity
                 ToggleChildrenChecked(idx, node, isChecked);
             }
 
-            ToggleParentFolderAndContainersChecked(idx, node, isChecked);
+            ToggleParentFolderAndContainersChecked(idx, node, checkState);
         }
 
         private void ToggleChildrenChecked(int idx, TNode node, bool isChecked)
@@ -270,8 +270,7 @@ namespace GitHub.Unity
             {
                 var childNode = Nodes[i];
 
-                var wasChecked = childNode.CheckState == CheckState.Checked;
-                SetCheckStateOnNode(node, isChecked);
+                SetCheckStateOnNode(childNode, isChecked);
 
                 if (childNode.IsFolderOrContainer)
                 {
@@ -327,8 +326,9 @@ namespace GitHub.Unity
             }
         }
 
-        private void ToggleParentFolderAndContainersChecked(int idx, TNode node, bool isChecked)
+        private void ToggleParentFolderAndContainersChecked(int idx, TNode node, CheckState checkState)
         {
+            var isChecked = checkState != CheckState.Empty;
             while (true)
             {
                 if (node.Level > 0)
@@ -381,9 +381,7 @@ namespace GitHub.Unity
 
                     var parentNodeState =
                         siblingsInSameState
-                            ? isChecked
-                                ? CheckState.Checked
-                                : CheckState.Empty
+                            ? node.CheckState
                             : CheckState.Mixed;
 
                     SetCheckStateOnNode(parentNode, parentNodeState);
--- a/src/GitHub.Api/UI/TreeBase.cs
+++ b/src/GitHub.Api/UI/TreeBase.cs
@@ -183,23 +183,9 @@ namespace GitHub.Unity
 
         public void SetCheckStateOnAll(bool isChecked)
         {
-            var nodeCheckState = isChecked ? CheckState.Checked : CheckState.Empty;
             foreach (var node in Nodes)
             {
-                var wasChecked = node.CheckState == CheckState.Checked;
-                node.CheckState = nodeCheckState;
-
-                if (!node.IsFolder)
-                {
-                    if (isChecked && !wasChecked)
-                    {
-                        AddCheckedNode(node);
-                    }
-                    else if (!isChecked && wasChecked)
-                    {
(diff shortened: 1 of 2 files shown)


src/GitHub.Api/UI/TreeBase.cs
```diff
@@ -332,8 +300,34 @@ private List<TNode> GetLeafNodes(TNode node, int idx)
...
+
+            var wasChecked = node.CheckState == CheckState.Checked;
+
+            node.CheckState = setCheckState;
+
+            if (!node.IsFolder)
+            {
+                if (isChecked && !wasChecked)
+                {
+                    AddCheckedNode(node);
```
queerviolet: Out of curiosity, what do `AddCheckedNode` and `RemoveCheckedNode` do? Are they callbacks—like, "this node is being added/removed to/from the set of checked nodes?"

src/GitHub.Api/UI/TreeBase.cs
```diff
@@ -332,8 +300,34 @@ private List<TNode> GetLeafNodes(TNode node, int idx)
...
+                {
+                    RemoveCheckedNode(node);
+                }
+            }
+        }
+
+        private void ToggleParentFolderAndContainersChecked(int idx, TNode node, bool isChecked)
         {
             while (true)
             {
```
queerviolet: wow does it ever seem like there should be a simpler way to track a bunch of checkboxes.

(I'm not saying there is. Just that it seems like there ought to be.)

src/GitHub.Api/UI/TreeBase.cs
```diff
@@ -384,14 +378,15 @@ private void ToggleParentFoldersChecked(int idx, TNode node, bool isChecked)
...
-                    if (siblingsInSameState)
-                    {
-                        parentNode.CheckState = isChecked ? CheckState.Checked : CheckState.Empty;
-                    }
-                    else
-                    {
-                        parentNode.CheckState = CheckState.Mixed;
-                    }
+
+                    var parentNodeState =
```
queerviolet: I <3 nested ternaries.

src/tests/UnitTests/UI/TreeBaseTests.cs
```diff
@@ -606,6 +611,74 @@ public void ShouldPopulateTreeWithSingleEntryWithMetaInPath()
                 }
             });
         }
+
+        [Test]
+        public void ShouldCheckParentOfMetaFile()
```
queerviolet: I know these are a pain to set up, but could we maybe do a test where we have
  - [X] Root
    - [x] Parent
      - [X] A
      - [X] B
      - [x] C

And then we uncheck `C` and see `Parent` and `Root` both go grey?

StanleyGoldman:
@queerviolet 🏆. That was a good call. You made me find two bugs.

src/GitHub.Api/UI/TreeBase.cs
queerviolet: Yeah, I figured.

src/tests/UnitTests/UI/TreeBaseTests.cs
```diff
@@ -606,6 +611,196 @@ public void ShouldPopulateTreeWithSingleEntryWithMetaInPath()
...
+            testTree.ToggleNodeChecked(5, cNode);
+
+            // Checked c.txt, everything checked
+
+            Assert.AreEqual(CheckState.Checked, rootNode.CheckState);
+            Assert.AreEqual(CheckState.Checked, parentNode.CheckState);
+            Assert.AreEqual(CheckState.Checked, aNode.CheckState);
+            Assert.AreEqual(CheckState.Checked, bNode.CheckState);
+            Assert.AreEqual(CheckState.Checked, cNode.CheckState);
+
```
queerviolet: Can we do one more where we uncheck `a.txt`, `b.txt`, and `c.txt`, and verify that the parent unchecks entirely?

queerviolet: Actually, the whole tree should uncheck at that point, no? That feels like a really appropriate final assertion. ;)

queerviolet:
Yay! I have one more test request, then this looks good to go. 🙏

src/tests/UnitTests/UI/TreeBaseTests.cs
StanleyGoldman: No problem

PR #959, opened 2018-11-07
Report the correct editing status to Unity. Fixes #934

### LINKED ISSUE #934: Inspector disabled for locked files
When a file is locked like a material. The inspector view UI is disabled. Is there a way to stop this?

shana:
There's two issues going on here:

- The `loggedInUser` field wasn't getting seeded with user information on load, so checking whether the current user owns a lock would always fail until the connections changed event got triggered somehow

- `GetLock` didn't account for not having user information, so if the user was logged out, it would default to not allowing any edits. Since the info wasn't getting seeded, it would always report files as being locked by someone else. It now defaults to allowing edits if the user isn't signed in. Now, this isn't a bug, but if the user is not signed in, all bets are kinda off on validating locks. We should probably warn the user that we can't check whether they own the lock, but we should probably not lock down Unity entirely because of it.

### DIFF
--- a/src/UnityExtension/Assets/Editor/GitHub.Unity/UI/LfsLocksModificationProcessor.cs
+++ b/src/UnityExtension/Assets/Editor/GitHub.Unity/UI/LfsLocksModificationProcessor.cs
@@ -21,6 +21,8 @@ namespace GitHub.Unity
             environment = env;
             platform = plat;
             platform.Keychain.ConnectionsChanged += UserMayHaveChanged;
+            // we need to do this to get the initial user information up front
+            UserMayHaveChanged();
 
             repository = environment.Repository;
             if (repository != null)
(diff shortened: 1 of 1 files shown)


src/UnityExtension/Assets/Editor/GitHub.Unity/ExtensionLoader/UnityAPIWrapper.cs
```diff
@@ -0,0 +1,20 @@
...
+using System;
+
+namespace GitHub.Unity
+{
+    [InitializeOnLoad]
+    public class UnityAPIWrapper : ScriptableSingleton<UnityAPIWrapper>
+    {
+        static UnityAPIWrapper()
+        {
+#if UNITY_2018_2_OR_NEWER
```
StanleyGoldman: Where or how is this preprocessor directive set?

src/UnityExtension/Assets/Editor/GitHub.Unity/UI/ProjectWindowInterface.cs
```diff
@@ -222,7 +222,8 @@ private static void OnLocksUpdate()
                 guidsLocks.Add(g);
             }
 
-            EditorApplication.RepaintProjectWindow();
+            //EditorApplication.RepaintProjectWindow();
+            UnityEditorInternal.InternalEditorUtility.RepaintAllViews();
```
StanleyGoldman: Can you tell me a bit more about this function call?

shana: We need to repaint not only the project window, but also the inspector, so that we can show the "this thing is locked by X" and that the `IsOpenForEdit` call happens and the inspector is disabled. There's no way to refresh the editor directly (well, there is, but it's an internal api), so this just causes Unity to repaint everything. Nail, meet bazooka, unfortunately, but that's the only way to do it with public APIs ¯\_(ツ)_/¯

StanleyGoldman: This seems important and something that we can also forget why.
I think you should add that explanation as a comment.

StanleyGoldman: Add comment to explain the rational here...
https://github.com/github-for-unity/Unity/pull/959#discussion_r236694800

src/UnityExtension/Assets/Editor/GitHub.Unity/ExtensionLoader/UnityAPIWrapper.cs
shana: That's Unity setting it, every Unity version sets defines so you can conditionally compile stuff

StanleyGoldman: I'm still confused, I thought preprocessor directives get "compiled out". So if we compile and distribute with unity 5.6, and the directive stays (?) then Unity 2018.2 recompiles on first load with the flag set?

StanleyGoldman: Okay, so to answer my own question. Yes, of course preprocessor directives get compiled out.

The contents of `ExtensionLoader` are distributed directly, not compiled. So the `.cs` files get compiled in the target Unity instance where the preprocessor directives get set as expected.

StanleyGoldman:
Okay, I understand all of this now. I'll do some tests against the newer version of Unity and I'll be ready to merge this.

Issue #948, opened 2018-10-26
History of a single file


sambonfire:
I'd like to be able to use the client to view the history of a single file; ie:

`git log -- <file>`

StanleyGoldman:
Hey @sambonfire. Thanks for checking our stuff out.
Yep. We definitely would like to get to all these smaller git features. Half the problem is figuring out how to put it in the UI without being an eyesore. We'll update this issue when we start to give it a shot.

sambonfire:
It would be wonderful if we could just start integrating these into the client API.  We're designing a workflow a bit different than what you guys have envisioned for the plugin, but we'd really like to share the same API layer.  Hopefully we could just add these to the API without worrying about UX immediately?

StanleyGoldman:
Hey @sambonfire that is a reasonable request.

When I get some time I can give it a shot, but I'm working on quite a few different projects at the same time.

If you don't getting it started. I will definitely help you get it done and :shipit:.
I will even send you a tshirt and some stickers for your effort. _**(I am not above bribery)**_ 😄

We have documentation on getting started with building the project.
If you have any questions ask away and I'll get back to you.

sambonfire:
I'm totally not above bribery either! :)

I wasn't able to get the code to build and pass tests on my mac; i'll give it a shot on my windows box.  expect a PR.

StanleyGoldman:
If you wanna set aside 30 minutes or so, I don't mind giving you a tour of the things you'd need to change to get what you want. I'm in EST.

StanleyGoldman:
Implemented in #999

Issue #947, opened 2018-10-26
Checkout a specific version of a file(s)


sambonfire:
This is a feature request.  I'd like the client to support checking out a set of files at a specific changeset.

This is the equivalent of the command:

`git checkout <changset> -- <list of files>`

StanleyGoldman:
Hey @sambonfire. This is another one for the figuring out.

sambonfire:
Similar to my last comment, it would be wonderful if we could just start integrating these into the client API so we can use it in our workflows that we are designing.

StanleyGoldman:
Hey @sambonfire same from me again. If you can start it I will chip in and help.

StanleyGoldman:
Implemented in #999

PR #961, opened 2018-11-09
File History

sambonfire:
### Description of the Change

I'm submitting this PR right now mostly for feedback; there are a handful of things that I'd still like to fix/cleanup; I just want early feedback!

I wanted to be able to display a window showing the history of a single file.  On this file history view, users should be able to "check out" an old version of the file.  This is a common theme in game development -- just wanting to roll back one file to test vs. all files in a change.

In order to accomplish this, 2 new commands were added to the git client:
 - log history of a file
 - checkout a set of files at a specific change

I also made some interfaces public; this was mostly because I initially started developing the user interface outside of the github plugin because I wasn't sure you'd want it.  I've left these changes in this PR because I think that they'll be worthwhile in the future.

To view file history, users can select files in the project file explorer, right click on them, then select Git -> History.  This pops up a new window showing the history of the selected file(s).

### Alternate Designs

A new view could have been built into the main github window.

### Benefits

Users can quickly view history of a single file and revert to any point in history with just a few clicks.

### Possible Drawbacks

Currently, there is no checks to see if the file has modifications.  As is, the checkout is destructive and this could lead to data loss.  I plan on fixing this by adding a confirmation dialog.

### Applicable Issues

#948, #947


StanleyGoldman:
I haven't forgotten you. 💛

StanleyGoldman:
I'm working on getting everything that was in master out the door into a release. Then we can start working on this together for the next release. Sorry to keep you hanging.

sambonfire:
No problem!

sam

StanleyGoldman:
Happy New Year. This is on my radar this week.

src/GitHub.Api/Application/ApiClient.cs
```diff
@@ -482,7 +482,7 @@ class GitHubUser
         public string Login { get; set; }
     }
 
-    class GitHubRepository
+    public class GitHubRepository
```
StanleyGoldman: If I'm not mistaken this class can remain internal.

src/GitHub.Api/Git/GitClient.cs
```diff
@@ -208,6 +208,15 @@ public interface IGitClient
...
         ITask<string> DiscardAll(IOutputProcessor<string> processor = null);
 
+        /// <summary>
+        /// Executes at least one `git checkout` command to checkout files at the given changeset
+        /// </summary>
+        /// <param name="changeset">The md5 of the changeset</param>
+        /// <param name="files">The files to check out</param>
+        /// <param name="processor">A custom output processor instance</param>
+        /// <returns>String output of git command</returns>
+        ITask<string> CheckoutVersion(string changeset, IEnumerable<string> files, IOutputProcessor<string> processor = null);
```
StanleyGoldman: For the sake of immediate consistency we should probably take `IList<string>` here.
That being said, I'm not sure why we use `IList<string>`. I'll log a TODO issue for this.

src/GitHub.Api/Application/ApiClient.cs
StanleyGoldman: I'm wrong.

StanleyGoldman:
I sent you a pull request to fix the build

https://github.com/bonfirestudios/Unity/pull/1

StanleyGoldman:
After that pull request is merged in...

Adding some screenshots for further discussion

image

image

StanleyGoldman:
@donokuda put this mockup together for us. I think I like it.

image

So we are going to make a "Pinnable" File Log History view.

donokuda:
👋Adding a little more clarity to what I was thinking with my mock-up. For the most part, everything that is already built is kept as is (including having the file history appear as a separate window.)

The main thing was making the header feel more like the Unity inspector where we show the file's large icon in the header as well as the filename. Clicking the "Show in Project" button would highlight the file in the project pane (which is the current functionality of clicking the "GIT File history for [filepath]" link) 

❓Something I'm not 100% sure about is whether it makes sense to show the full path or just the filename. The benefit for the full path is that it ensures that we're looking at the right file and where it's located. The downside is that we can run out of space quickly. Right now, I'm leaning towards showing just the file name since the "Show in Project" button could also be a compromise that's already built.

sambonfire:
<3 the mockup and ideas (dockable history view).  I should have some time in the next couple weeks to work on this!

sam

sambonfire:
Added some code here that will see if the file has any local changes, and if so it prompts the user to overwrite it (when doing a Checkout on the revision).  I also fixed some bugs where the UI was causing exceptions after a domain reload -- it should behave now.

Are there any other outstanding changes you'd like me to make?

StanleyGoldman:
Hey @sambonfire I'm going to close this pull request, but I'm going to continue to graft whatever remaining functionality you added to the new pull request #999 

I don't think you will be able to push to that branch, but you can open additional pull requests adding functionality to that branch.

PR #973, opened 2018-11-29
configure: UnityYamlMerge as a merge tool

### LINKED ISSUE #943: Is UnityYAMLMerge actually being used?
### Description

While investigating using this plugin for our next project I noticed that the UnityYAMLMerge tool seems to get configured for use but never actually gets launched.

When creating a project with the plugin the following configuration options get set:

_.gitattributes_:
```
# Unity files
*.meta -text merge=unityamlmerge diff
*.unity -text merge=unityamlmerge diff
*.asset -text merge=unityamlmerge diff
*.prefab -text merge=unityamlmerge diff
```

_.git/config_:
```
[merge "unityyamlmerge"]
	cmd = 'C:\\Program Files\\Unity 2018.1.1f1\\Editor\\Data\\Tools\\UnityYAMLMerge.exe' merge -p $BASE $REMOTE $LOCAL $MERGED
	trustExitCode = false
```

This looks well and good but:
- Attributes file misspells the tool as "unityamlmerge" (note the missing "y").
- Config file doesn't misspell the tool name.
- Even if the names did match the tool still wouldn't run because the `merge` attribute looks for a custom low-level merge driver in the config file in the following format:
```
[merge "mergeDriverName"]
	name = Human-readable name
	driver = mergeDriverExecutable %O %A %B %L %P
	recursive = binary
```
- The config file as is uses the **custom merge driver** header `[merge "unityyamlmerge"]` but **external merge tool** fields: `cmd =` `trustExitCode =`. These are different things but appear to be mixed here. A `mergetool` section is supposed to look like this:
```
[mergetool "extMerge"]
  cmd = extMerge "$BASE" "$LOCAL" "$REMOTE" "$MERGED"
  trustExitCode = false
```
But for that to work it also requires this entry, which is missing:
```
[merge]
  tool = extMerge
```
And I think that would launch the tool for all file types not just the select ones.

**Disclaimer:**
I'm not very proficient in git so I might be missing something here but this is what it looks like to me after investigating the code and docs.

I'm using **GitHub for Unity 1.1.0** and **Unity 2018.1.1f1**.

### References

https://git-scm.com/book/en/v2/Customizing-Git-Git-Configuration#_external_merge_tools
https://git-scm.com/docs/gitattributes#_defining_a_custom_merge_driver
https://git-scm.com/docs/git-config#git-config-mergetool
https://docs.unity3d.com/Manual/SmartMerge.html

https://github.com/github-for-unity/Unity/blob/1c53148a39f81d3146b3843eab21efef16115306/src/GitHub.Api/Resources/.gitattributes#L3-L7

https://github.com/github-for-unity/Unity/blob/1c53148a39f81d3146b3843eab21efef16115306/src/GitHub.Api/Application/ApplicationManagerBase.cs#L291-L299

bk2204:
### Description of the Change

UnityYamlMerge is a merge tool which can help with conflict resolution.
However, currently we configure it using the merge section instead of
the mergetool section, meaning that it's not being used.  In addition,
we configure it in .gitattributes, even though that has no effect.

Configure this tool in the correct section of the Git config file and
skip inserting it into the gitattributes file.

This should fix #943.

Please note that I have not actually tested this change, as I don't have Unity or any of the tooling required to test this; it's merely been submitted to address a question that @StanleyGoldman brought to my attention.  Testing it before merge would be prudent.

### Alternate Designs

I considered whether to set this up as a merge driver instead, but it doesn't appear that that's a valid configuration for UnityYamlMerge.  If it turns out it is, it's easy to add in addition. 

### Benefits

UnityYamlMerge will be usable as a merge tool.

### Possible Drawbacks

There shouldn't be any.  The tool didn't work before, and it should work now.

I suppose users could be unhappy with the merge tool and wish to go back to the way things were before, but that's a risk with any change.

### Applicable Issues

#943

### DIFF
--- a/src/GitHub.Api/Application/ApplicationManagerBase.cs
+++ b/src/GitHub.Api/Application/ApplicationManagerBase.cs
@@ -288,12 +288,12 @@ namespace GitHub.Unity
                 ? $@"'{unityYamlMergeExec}' merge -p ""$BASE"" ""$REMOTE"" ""$LOCAL"" ""$MERGED"""
                 : $@"'{unityYamlMergeExec}' merge -p '$BASE' '$REMOTE' '$LOCAL' '$MERGED'";
 
-            GitClient.SetConfig("merge.unityyamlmerge.cmd", yamlMergeCommand, GitConfigSource.Local).Catch(e => {
+            GitClient.SetConfig("mergetool.unityyamlmerge.cmd", yamlMergeCommand, GitConfigSource.Local).Catch(e => {
                 Logger.Error(e, "Error setting merge.unityyamlmerge.cmd");
                 return true;
             }).RunSynchronously();
 
-            GitClient.SetConfig("merge.unityyamlmerge.trustExitCode", "false", GitConfigSource.Local).Catch(e => {
+            GitClient.SetConfig("mergetool.unityyamlmerge.trustExitCode", "false", GitConfigSource.Local).Catch(e => {
                 Logger.Error(e, "Error setting merge.unityyamlmerge.trustExitCode");
                 return true;
             }).RunSynchronously();
--- a/src/GitHub.Api/Resources/.gitattributes
+++ b/src/GitHub.Api/Resources/.gitattributes
@@ -1,10 +1,10 @@
 * text=auto
 
 # Unity files
-*.meta -text merge=unityamlmerge diff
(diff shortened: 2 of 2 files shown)


StanleyGoldman:
Thanks @bk2204

PR #974, opened 2018-11-29
Fixing UnityYamlMerge

StanleyGoldman:


shana:
We're going to need to also fix existing installations...

PR #976, opened 2018-11-30
Capitalize "ok"

FurkanKambay:
Just bothered me; sorry for an atomic PR... 😅 I might contribute something real at a later time.
PS I really wanted to fill the template for sarcasm reasons but I got lazy.


StanleyGoldman:
LOL, thanks for the laugh and the fix.

Let me know when you wanna try something more, I'm always down to help.

PR #978, opened 2018-12-03
Fixing UnityYamlMerge to use merge driver

### LINKED ISSUE #943: Is UnityYAMLMerge actually being used?

StanleyGoldman:
Related to #974 
Fixes #943 

1. On project initialization, configure UnityYamlMerge to be executed as a merge driver
1. On first load of a projeect, if a repository exists.
    1. Remove any UnityYamlMerge merge tool configuration
    1. If the incorrectly spelled legacy `unityamlmerge` is found in the `.gitattributes` file
        1. Add an additional misspelled configuration for merge driver configuration
        1. Remove any misspelled merge tool configurations


StanleyGoldman:
Thanks @bk2204 for teaching us how merge drivers are supposed to work

PR #989, opened 2018-12-21
Making the standalone api nicer to use

shana:
I did a little app (https://github.com/shana/unity-runtime-git) to test using the api at runtime, and I had to do some tweaks to make it nicer to use. These are those tweaks:

- When parsing the history log, the first line of the commit was being added to the description. This makes it annoying to show entries when showing both the first line (summary) and the body (description), as the first line would appear duplicated. I kinda wanna treat these as email subject and body (because that's what they were meant to be originally in git anyways), so took out the duplication.
- calling `InitializeRepository(path to repository)` on `DefaultEnvironment` does the useful thing of setting the base path on the FileSystem object, which means any shell commands and path manipulations are going to assume that path to repository as the base for all relative paths and working directory paths. Which is handy! Only the logic that we currently have makes it impossible to call `InitializeRepository` twice (it just ignores it), so it's not easy to switch these paths.
- The task classes should be public, they're useful!

### DIFF
--- a/src/GitHub.Api/OutputProcessors/LogEntryOutputProcessor.cs
+++ b/src/GitHub.Api/OutputProcessors/LogEntryOutputProcessor.cs
@@ -143,7 +143,6 @@ namespace GitHub.Unity
                         }
 
                         summary = line;
-                        descriptionLines.Add(line);
                         phase++;
                         // there's no description so skip it
                         if (oneliner)
--- a/src/GitHub.Api/Platform/DefaultEnvironment.cs
+++ b/src/GitHub.Api/Platform/DefaultEnvironment.cs
@@ -83,7 +83,7 @@ namespace GitHub.Unity
             Guard.NotNull(this, FileSystem, nameof(FileSystem));
 
             NPath expectedRepositoryPath;
-            if (!RepositoryPath.IsInitialized)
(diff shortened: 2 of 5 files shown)


src/tests/UnitTests/IO/LogEntryOutputProcessorTests.cs
```diff
@@ -61,6 +61,49 @@ public void ShouldParseSingleCommit()
...
+
+            var commitTime = new DateTimeOffset(2017, 1, 6, 15, 36, 57, TimeSpan.FromHours(1));
+
+            var expected = new[]
+            {
+                new GitLogEntry("1cd4b9154a88bc8c7b09cb8cacc79bf1d5bde8cf",
+                    "Author Person", "author@example.com",
+                    "Author Person", "author@example.com",
+                    "Rename RepositoryModelBase to RepositoryModel",
+                    Environment.NewLine + "This is a line on the description",
```
StanleyGoldman: Wouldn't we want to remove the newline from the Body?

shana: I'm not sure if the GitLogEntry should be trim whitespace when parsing things, but maybe it should, yeah

StanleyGoldman: Did this change?

PR #992, opened 2018-12-28
Update the generated .gitignore

FurkanKambay:
### Description of the Change

Updated the `.gitignore` that gets generated when initializing the repo. I copied from [Unity.gitignore in github/gitignore](https://github.com/github/gitignore/blob/master/Unity.gitignore).

To go over the changes top-to-bottom:
- Removed the leading slashes since the `.gitignore` is at the root of the repo anyway
- Ignored `Logs/` (github/gitignore#2912)
- Ignored `.gradle/` (github/gitignore#2906)
- Ignored `*.opendb` (github/gitignore#2681)
- Ignored `*.VC.db` (github/gitignore#2847)
- Ignored `*.pdb.meta` (github/gitignore#2486)
- Fixed casing ([😛](/github/gitignore/pull/2900))

Also there was [another change](/github/gitignore/pull/2874) about Crashlytics, but I didn't include it. Tell me if you want it.


src/GitHub.Api/Resources/.gitignore
```diff
@@ -1,13 +1,16 @@
...
-/[Oo]bj/
-/[Bb]uild/
-/[Bb]uilds/
+[Ll]ibrary/
+[Tt]emp/
+[Oo]bj/
+[Bb]uild/
+[Bb]uilds/
+[Ll]ogs/
+Assets/AssetStoreTools*
```
FurkanKambay: Anyone know why this wildcard `*` is here? I only see one folder for that asset.
```suggestion
Assets/AssetStoreTools/
```

FurkanKambay: Right. Forgot about the meta file, thanks.

shana: You know what, you have a point, maybe it is better not to have this ignore here, it's going to be confusing for people if they want to commit the tooling into their repo and it's ignored by default.

shana:
This looks great, thank you so much! Sorry for the delay in the review, the holidays always get in the way...

I did a couple of tweaks to your branch, everything looks great. As for the crashalytics line, I don't know what it is, but looking at examples like https://github.com/auth0/sharelock-android/blob/master/app/src/main/assets/crashlytics-build.properties on the internets, looks like it's something that shouldn't be committed to source control. It's not always in StreamingAssets though, so the ignore should probably just be the filename `crashlytics-build.properties`, I've made that change here. The gitignore repo needs a fix :P

FurkanKambay:
@shana No worries! And thanks for the other changes.

PR #998, opened 2019-01-08
Add missing YAML files to .gitattributes

### LINKED ISSUE #847: Add `diff` attribute to .gitattributes for text assets
We should add a `diff` attribute to `.gitattributes` on the filetypes that are text files (like prefab, asset, unity, etc), so that diffs for them are shown on the command line (currently git skips showing diff because it thinks these files are binary)

FurkanKambay:
Fixes #847 

(The generated) .gitattributes file was missing some other YAML files besides `.prefab`, `.unity`, etc.

I created an empty project and created almost everything under `Assets`/`Create`. Here are the files: https://i.imgur.com/QODfnOr.png

While I'm not familiar with a lot of these extensions, Unity supports them, so they should be in .gitattributes.

### DIFF
--- a/src/GitHub.Api/Resources/.gitattributes
+++ b/src/GitHub.Api/Resources/.gitattributes
@@ -6,6 +6,23 @@
 *.asset -text merge=unityyamlmerge diff
 *.prefab -text merge=unityyamlmerge diff
 *.mat -text merge=unityyamlmerge diff
+*.anim -text merge=unityyamlmerge diff
+*.controller -text merge=unityyamlmerge diff
+*.overrideController -text merge=unityyamlmerge diff
+*.physicMaterial -text merge=unityyamlmerge diff
+*.physicsMaterial2D -text merge=unityyamlmerge diff
+*.playable -text merge=unityyamlmerge diff
+*.mask -text merge=unityyamlmerge diff
+*.brush -text merge=unityyamlmerge diff
+*.flare -text merge=unityyamlmerge diff
+*.fontsettings -text merge=unityyamlmerge diff
+*.guiskin -text merge=unityyamlmerge diff
+*.giparams -text merge=unityyamlmerge diff
+*.renderTexture -text merge=unityyamlmerge diff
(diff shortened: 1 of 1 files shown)


shana:
Cooooooooooooooool 😃 😃 😄 !

There are more (or so Unity peeps tell me), but the only way to know is to scrape the Unity source (or so Unity peeps tell me), so yay this is a great start!

shana (approved):

FurkanKambay:
@shana
Hey! (IDK if I should post this on a closed PR but I didn't want to open an issue for this)

So, I kinda took a look at the Unity source code 😄  and found something.
I think this is kind of a central location for the file extensions:
[switch statement in `FindIconForFile()`](https://github.com/Unity-Technologies/UnityCsReference/blob/0cf7cdad8092152ba078b4e306b4182608de040d/Editor/Mono/InternalEditorUtility.cs#L25)

I extracted all the extensions in the `switch` statement and categorized them. I also added [the template in alexkaratarakis/gitattributes](https://github.com/alexkaratarakis/gitattributes/blob/1a244a639484cf55209aeede8d7fe6cec4f52ea9/Unity.gitattributes) to the equation.

Here's a JSON file that has the differences between all 3 repos: [gist](https://gist.github.com/FurkanKambay/228939374abd6a9f79c4486690cf3c0d)
(I recommend folding everything to see the structure)

Now, if we look at just the Unity-specific extensions that we don't have here, the list is short:
`asmdef, asmref, uss, uxml, prefs`.

And these ones in the template repo: `cubemap, unitypackage` (I think the other 2 aren't valid extensions)

Otherwise, there is quite a lot of difference...

In the Unity source code, there are more extensions for file types like images, video files, etc. We can ignore them or add some of them. (Not all of them are commonly used)

Also, there are extensions that we have but the Unity source code doesn't contain, which is concerning 😟

Anyway, do you think adding any of these extensions is worthwhile? If so, should I open a PR here or in the new repo?

Thanks!

PR #999, opened 2019-01-17
File History Window

StanleyGoldman:
Building upon work started in #948 by @sambonfire


sambonfire:
holy crap this is looking awesome.

StanleyGoldman:
image

> @donokuda: https://github.com/github-for-unity/Unity/pull/961#issuecomment-454105771
>
> 👋Adding a little more clarity to what I was thinking with my mock-up. For the most part, everything that is already built is kept as is (including having the file history appear as a separate window.)
> 
> The main thing was making the header feel more like the Unity inspector where we show the file's large icon in the header as well as the filename. Clicking the "Show in Project" button would highlight the file in the project pane (which is the current functionality of clicking the "GIT File history for [filepath]" link) 
> 
> ❓Something I'm not 100% sure about is whether it makes sense to show the full path or just the filename. The benefit for the full path is that it ensures that we're looking at the right file and where it's located. The downside is that we can run out of space quickly. Right now, I'm leaning towards showing just the file name since the "Show in Project" button could also be a compromise that's already built.

StanleyGoldman:
Some action shots

donokuda:
😍This is looking so good and thanks for posting a gif of it in action! The only thing I can pick out is that the "Show in Project" button gets bumped to outside of the interface:

screen shot 2019-02-12 at 9 52 49 am

I'm assuming that it has something to do with the [`FlexibleSpace()` pushing the `BeginVerticalGroup()`](https://github.com/github-for-unity/Unity/blob/299890befcfbd37a7d06e45c6a64b41e62067947/src/UnityExtension/Assets/Editor/GitHub.Unity/UI/FileHistoryWindow.cs#L231-L242) to the side.

I took an attempt at re-writing the `DoHeaderGUI()` method with something that _might_ fix the button issue, except I haven't tested it myself locally because it's been [well over a year](https://github.com/github-for-unity/Unity/commits?author=donokuda) since I've touched the project:

```csharp
private void DoHeaderGUI() {
  GUILayout.BeginHorizontal(Styles.HeaderBoxStyle);
  {
    var iconWidth = 32;
    var iconHeight = 32;

    GUILayout.Label(selectedIcon, GUILayout.Height(iconWidth), GUILayout.Width(iconHeight));
    GUILayout.Space(16);

    GUILayout.BeginVertical();
    {
      GUILayout.Label(selectedObjectAssetPath, Styles.FileHistoryLogTitleStyle);

      GUILayout.BeginHorizontal();
      {
        GUILayout.FlexibleSpace();

        if (GUILayout.Button("Show in Project"))
        {
            EditorGUIUtility.PingObject(selectedObject);
        }
      }
      GUILayout.EndHorizontal();
    }
    GUILayout.EndVertical();
  }
  GUILayout.EndHorizontal();
}
```

Would anyone be able to push a tweak for this? If not, I can take this on, but I'll need some time to get my environment set up again.

StanleyGoldman:
@donokuda as always you work miracles. I also made sure you get credit for that code.

src/UnityExtension/Assets/Editor/GitHub.Unity/UI/ContextMenu.cs
```diff
@@ -0,0 +1,12 @@
+﻿using System.Collections;
+using System.Collections.Generic;
+using UnityEngine;
+using UnityEditor;
+
+namespace GitHub.Unity
+{
+    public class ContextMenu
```
shana: What is this one for?

src/UnityExtension/Assets/Editor/GitHub.Unity/UI/FileHistoryWindow.cs
```diff
@@ -0,0 +1,250 @@
...
+            AttachHandlers(Repository);
+
+            if (HasRepository)
+            {
+                
+            }
+            else
+            {
+               
+            }
```
shana: This if block had something at some point, I'm guessing?

src/UnityExtension/Assets/Editor/GitHub.Unity/UI/ContextMenu.cs
StanleyGoldman: 👍

src/UnityExtension/Assets/Editor/GitHub.Unity/UI/FileHistoryWindow.cs
StanleyGoldman: 👍

StanleyGoldman: Thanks @shana

StanleyGoldman:
@sambonfire let me know if there is anything you'd like to change about this before it goes in.

meaghanlewis:
This functionality looks good to me @StanleyGoldman 👍

PR #1008, opened 2019-01-30
refresh asset database after revert

sambonfire:
### Description of the Change

Refresh asset database after reverting.  This is necessary to make sure Unity is looking at the most up-to-date files.  I didn't see that this was happening in this case so it's a pretty simple change.

### Benefits

Users see accurate versions of file in Unity

### Possible Drawbacks

It is possible it is a redundant call, but not from what I could tell.


StanleyGoldman:
Thanks @sambonfire I'm working to land this for a release coming soon.

PR #1014, opened 2019-02-12
Sometimes .tiff images has .tif extension

terokorp:
even on asset store

### Description of the Change
Adding .tif (alternative extension for .tiff) files to LFS at default

### Benefits
Others dose't do same mistake that I just did; commited over 2G non lfs git


StanleyGoldman:
Hey @terokorp 

Yes, you are right about that. Thanks for the contribution.

PR #1058, opened 2019-06-14
Fix the git log and git status output processors. Split status field into two.

shana:
The git status output processor was not handling all the variations that the command outputs for the status flags. Make it actually follow what the manual says.

In the process, the status field of a git status entry has now been split into indexStatus and workTreeStatus fields, to provide all the underlying information that the command provides.

Staged and Unmerged can be obtained via new properties in the GitStatusEntry structure, or calculated directly from the indexStatus and workTreeStatus fields.

Also bump version to 1.4, since there's new API and stuff


sna-scourtney:
We really appreciate this fix, as we've been diagnosing some problems with remote repos. I do unfortunately have to report a problem related to this, though. In testing, I attempted to pull a remote repo where there were a lot of file conflicts with the local. The popup error message now provides a great deal of useful information, but unfortunately the window can be so large that its title bar and buttons are off the screen, making it difficult or impossible to move, resize, or dismiss it. (There may be a Windows hotkey to do this, but I'm not aware of it.) The net result is that I'm forced to use Task Manager to kill Unity Editor (no big deal in this testing sandbox, but still a problem for real projects).

I am attaching a screen capture of the problem. I run on a 4K screen, so in reducing this down to a reasonable size it may not be easily readable. The point is just the way the window is positioned. My recommendation would be to put the command output into a scrollable text area widget and then allow the overall popup window to be resized by the usual method of dragging its corner. In a perfect world, a checkbox would toggle word wrap for the text area, but that's not essential as long as it could be copied to the clipboard and pasted into a text editor.

Thanks again for the fix. It's a big step in the right direction, and I'm sorry to have to report an issue so quickly.
GHU-WindowOverflow-20190620-s

Issue #1068, opened 2019-08-02
Constant NullReferenceException in GitHub.Unity.Window.Update()


DreadKyller:
### Prerequisites

Trace logging is enabled, here is the log file contents:
```
190801-21:46:36.726  INFO [ 1] <Global>                            Initializing GitHubForUnity:'v1.4.0' Unity:'v2019.2.0f1'
190801-21:46:36.764 TRACE [150] <DownloadTask>                      Download of http://github-vs.s3.amazonaws.com/unity/latest.json to C:\Users\user\AppData\Local\GitHubUnity\latest.json Attempt 1 of 1
190801-21:46:36.910 TRACE [149] <ProcessWrapper>                    Running 'C:\Users\user\AppData\Local\GitHubUnity\git\cmd\git.exe --version'
190801-21:46:39.417 TRACE [149] <ProcessWrapper>                    Running 'C:\Users\user\AppData\Local\GitHubUnity\git-lfs\git-lfs.exe version'
190801-21:46:39.848 TRACE [150] <DownloadTask>                      Downloading http://github-vs.s3.amazonaws.com/unity/latest.json
190801-21:46:40.071 TRACE [149] <ProcessWrapper>                    Running 'C:\Users\user\AppData\Local\GitHubUnity\git\cmd\git.exe config --local --unset merge.unityyamlmerge.cmd'
190801-21:46:41.159 TRACE [149] <ProcessWrapper>                    Running 'C:\Users\user\AppData\Local\GitHubUnity\git\cmd\git.exe config --local --unset merge.unityyamlmerge.trustExitCode'
190801-21:46:41.450 TRACE [149] <ProcessWrapper>                    Running 'C:\Users\user\AppData\Local\GitHubUnity\git\cmd\git.exe config --replace-all --local merge.unityyamlmerge.name "Unity SmartMerge (UnityYamlMerge)"'
190801-21:46:41.456 TRACE [150] <DownloadTask>                      Downloading http://github-vs.s3.amazonaws.com/unity/latest.json StatusCode:200
190801-21:46:41.579 TRACE [150] <Global>                            Skipping GitHub for Unity update v1.4.0, we already have it
[...]

Issue #1053, opened 2019-05-28
Lazy<T> type existing in two namespaces


stonecompass:
### Prerequisites

- Newest HDRP package (5.13.0)
- Newest Github for Unity asset

### Description

There is an issue with the Lazy<T> type existing in two namespaces.
"Library\PackageCache\com.unity.render-pipelines.core@5.13.0\Runtime\Volume\VolumeManager.cs(13,25): error CS0433: The type 'Lazy<T>' exists in both 'System.Threading, Version=1.0.3333.0, Culture=neutral, PublicKeyToken=402899b480e6f383' and 'mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089'
" 

### Steps to Reproduce
1. Install HDRP
2. Install Github for Unity

**Expected behavior:** No compilation errors

**Actual behavior:** Getting a namespace issue for Lazy<T>

**Reproduces how often:** 100 %

shana:
@stonecompass  I assume this is a .net 4.5 project? All the DLLs are shipped disabled by default, and during loading it will enable the 2.0 or 4.5 versions depending on which runtime your project is targeting. For some reason, it seems that the System.Threading.DLL is getting enabled when it shouldn't be (it's only meant to be enabled for the 2.0 profile). GhU will soon be converted into a package targeting only the .net 4.5 profile, so this will no longer be an issue. In the meantime, you can disable loading the System.Threading DLL in the inspector, that should fix this.

Issue #1018, opened 2019-02-21
Error importing package in Unity


ejgutierrez74:
I cant install latest package of github for unity:

I downloaded the latest version 1.2.1 and import as a custom package, i select to imort all the files but this message arises for every file.

Copying Temp/Export Package/48eb153816693401397241b99e75328a/d516f2a1bec6a9645a084ef8c9237132/asset to Assets\Plugins\GitHub\Editor\AsyncBridge.Net35.dll: 

Im on unity 2017.4.21 and Ubuntu 18.04

Thanks

DouglasUrner:
This also happens on the macOS 10.14.3 beta.

StanleyGoldman:
@ejgutierrez74 This is related to the Ubuntu issues as I mentioned in the other issue you created.

https://github.com/github-for-unity/Unity/issues/1015

@DouglasUrner If you can send over a log file. I would appreciate it.

ejgutierrez74:
@StanleyGoldman dont know if its the same issue, probably is different.... This one was that if you download the package and want to import it in your project, it gets the error said in the first post.

This is the first step to unity github to work, if its not installed / import it cant work ;)

The other error is a second step, once installed ( imported in your project), you configure your account and the remote repository, but it doesnt work as expected with the log attached.

Perhaps the bug, error has the same root, but from outside seem different. If you consider there is the same error close this issue. 

Thansk

dilas12345:
I am having similar issue, this time around in my case i can't get my packages, say "There are no packages" and at the bottom it says "Error refreshing packages". And I am working on Ubuntu OS

Issue #962, opened 2018-11-12
Add UPM package distribution option


tlaedre:
It would be great if there was a distribution option to get Github4Unity as a upm package. There's no real reason for the tool to live inside the Assets/ folder.

nickmccurdy:
It's also confusing to have the plugin itself show up in changes, it has a lot of binary files that you wouldn't normally want to commit.

nickmccurdy:
I recently reinstalled Unity to get the latest version and then installed the GitHub extension, and it shows up in changes every time I launch this Unity project.

StanleyGoldman:
You should be free to commit the changes, that is the intention.
It shouldn't show up again after you commit it.

JoNax97:
Hey there!
I'm going through the process of moving GithubForUnity to a package, for personal use. I would like to commit it when it's ready, but I have a few questions:

1- Would com.github.unity be the right package name? 
2- Should the UPM version live along with the Assets folder version? Or may it become a replacement going forward?
3- I have modified ExtensionLoader to support loading from packages folder, but the logic for it is currently driven by a const. Is this fine or should it be sorted at runtime?
4- Is there any potential issue I should look into?

shana:
@JoNax97 Hey! This work is already underway at https://github.com/Unity-Technologies/Git-for-Unity, and I'm planning on porting that back here this week. If you want to try it out, the master branch over there is already usable as a package, you can try the sample project in the `UnityProject` folder to see how it works.

shana:
https://github.com/github-for-unity/Unity/pull/1070 has all the changes needed for this, early builds are available in https://ci.appveyor.com/project/shana/git-for-unity

shana:
This project is not maintained anymore. The replacement is Git for Unity, at https://github.com/spoiledcat/git-for-unity. If you have GitHub for Unity in your project, remove it first and then install the [Git for Unity](https://github.com/spoiledcat/git-for-unity) package, see the README in the link for instructions.

PR #1070, opened 2019-08-13
Package Manager package support, .net 4, Unity 2019, Git for Unity, and more

shana:
This is a large refactor to reuse the api source in https://github.com/Unity-Technologies/Git-for-Unity (via a fork in https://github.com/spoiled/Git-for-Unity), so this repo will only concern itself with building the UI.

It also changes the build system to output Package Manager packages, with .net 35 (for Unity 5.6 and above targeting the .net 2.0 profile), and .net 471 (for 2018+ targeting the .net 4 profile). Because there's so many changes to how things are packaged, there's no longer a loader trying to guess which DLLs to load/unload based on the active profile. Instead, there are separate packages for each profile.

This also indirectly switches to using dugite git packages on all platforms, and adds asmdef to all projects.

Fixes #1070 #1068 #1053 #1018 #962 #456


Ike-Mac:
Any ETA when this will be merged to master?  I'm in the process of upgrading.   Thxs :)

